PR #2, opened 2019-01-08
Repo automation `CODEOWNERS`

mitchhentges:
My initial response to each question is:

* The benefit to being auto-tagged for reviews is worth this extra configuration
* I'd throw this in `.github/CODEOWNERS` since it's github-specific, and so it doesn't clog the root of the repo with visible files
* I'd assign the most-specific team to each project. Everything in `mozilla-mobile` will probably be `@mozilla-mobile/releng`, and the rest will be `@mozilla-releng/releng` (maybe in the future we'll break down `@mozilla-releng/releng` into more specific teams as well, who knows :man_shrugging: )


rfcs/0001-repo-automation-code-owners.md
```diff
@@ -0,0 +1,35 @@
...
+# Summary
+
+Leverage [Github "code owners"](https://help.github.com/articles/about-code-owners/) for product repositories
+(e.g.: [`focus-android`](https://github.com/mozilla-mobile/focus-android/)) so that releng is automatically assigned
+as a reviewer when automation-related files are changed.
+
+## Motivation
+
+Right now, it's possible for product-level changes in automation to happen without releng being notified. This will
+allow us to give insight to such changes and avoid potential build-breaking accidents.
```
escapewindow: I think having insight is good, as is limiting it to specific subteams. It might be nice to have the ability for people not in the subteam to subscribe, but I'm not sure if that's supported.

rfcs/0001-repo-automation-code-owners.md
```diff
@@ -0,0 +1,35 @@
...
+
+# Details
+
+We will use the [Github "code owners"](https://help.github.com/articles/about-code-owners/) feature to declare the
+`automation` (or `tools`) directory to be "owned by" releng. This will assign us as reviewers to changes to this code.
+However, we will configure this so that review from releng isn't _required_, so that we don't hinder any product
+team progress.
```
escapewindow: I think this is a good starting point: stay informed but don't have the process hinder self-serve progress. As this process matures we can adjust.

rfcs/0001-repo-automation-code-owners.md
```diff
@@ -0,0 +1,35 @@
...
+
+Example `CODEOWNERS` file:
+```
+/tools @mozilla-mobile/releng
+/automation @mozilla-mobile/releng
```
escapewindow: Are we informed if someone removes releng from `CODEOWNERS`?

rfcs/0001-repo-automation-code-owners.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+# Open Questions
+
+* [ ] This is more complicated than just configuring ownership for an entire repository. Is this tradeoff worth it?
```
escapewindow: If we want a real answer to this, one low-touch experiment could be for the releng-mobile team to all subscribe to the repo(s) in question and stay up on all emails. If this is too noisy and would lead to people ignoring these emails, likely the tradeoff will be worth it.

rfcs/0001-repo-automation-code-owners.md
```diff
@@ -0,0 +1,35 @@
...
+* [ ] Where should this file go? (root of repo, `.github/CODEOWNERS`, `docs/CODEOWNERS`)?
```
escapewindow: Placing it inside one of the releng-owned areas would be one potential solution to the question above (about knowing if this file is changing). Also possible for releng to own this file.

rfcs/0001-repo-automation-code-owners.md
```diff
@@ -0,0 +1,35 @@
...
+* [ ] We have two releng teams: @mozilla-mobile/releng and @mozilla-releng/releng. Should we assign the mobile releng team to releng projects, and other relevant projects to the larger releng team? Should we assign all to just the larger releng team?
```
escapewindow: I'm guessing we start with the more specific team, and adjust later if needed/wanted.

rfcs/0001-repo-automation-code-owners.md
escapewindow: similarly, we may care about `.taskcluster.yml` and other infra config files.

mitchhentges: We can add `/.github/CODEOWNERS` to the `CODEOWNERS` and we'll be pinged when it changes, good point. I'll add this to the RFC

rfcs/0001-repo-automation-code-owners.md
mitchhentges: I can confirm that I already "unwatched" all the mobile repositories because I got a lot of noise

rfcs/0001-repo-automation-code-owners.md
mitchhentges: [There's only three locations that we can place this file](https://help.github.com/articles/about-code-owners/#codeowners-file-location), we can't put it in a releng-owned area (besides, it shouldn't be owned by _only_ us, all teams working in the repo care about `CODEOWNERS`)

rfcs/0001-repo-automation-code-owners.md
mitchhentges: Yeah, you can set individual people as owners as well (and you can have the same file/directory be owned by multiple people)

rfcs/0001-repo-automation-code-owners.md
escapewindow: Would it make sense for releng to own `.github/`? Meaning changes would notify us without blocking changes?

JohanLorenzo: I'd be in favor of using `.github/CODEOWNERS` as well because : 

* Some mobile projects already use the `.github` folder https://github.com/mozilla-mobile/fenix/tree/master/.github/ISSUE_TEMPLATE https://github.com/mozilla-mobile/focus-android/tree/master/.github/ISSUE_TEMPLATE
* This feature relies on a single tool: Github

mitchhentges:
[Rendered RFC](https://github.com/mitchhentges/releng-rfcs/blob/0001/rfcs/2-repo-automation-code-owners.md)
Moving to "Proposal" as all questions are answered.

@mozilla-releng/releng 
(note: I'm currently balancing noise vs. shopping RFCs around. Let me know if this ping is too broad)

PR #5, opened 2019-01-11
Shippable Builds

Callek:


rfcs/0005-shippable-builds.md
```diff
@@ -0,0 +1,47 @@
+# RFC 5 - Shippable Builds
```
escapewindow: Nit: I think @mitchhentges has been making RFCs start from 1 and increment, even if the number doesn't match the PR number. We've skipped 3 and 4 and have jumped to 5, I assume because it matches the PR number here. I don't have strong opinions which direction we go, but it's probably best if we are consistent :)

It does look like [you're following directions](https://github.com/mozilla-releng/releng-rfcs/blob/master/mechanics.md#make-a-pr).

rfcs/0005-shippable-builds.md
```diff
@@ -0,0 +1,47 @@
...
+* Comments: [#5](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/5)
+* Proposed by: @Callek
+
+# Summary
+
+Consolidate multiple build types and tests derived from those types into one solid answer for "is
+this build/push suitable to ship to our users". And to maintain a build/test platform, where we can
```
escapewindow: Rambling aside: I've struggled a bit with how to define shippable builds simply.

I used the question `Are these binaries shippable?` in the [original writeup](https://groups.google.com/forum/#!msg/mozilla.release.engineering/MRfHnhaVI48/l3kew8UmAAAJ), but a) that uses the term to define the term, and b) there are more criteria that go into the question than just the build configuration. (Is the code revision the one want to ship? Is it signed properly for the appropriate shipping channel? Is it built off of a shipping branch, or is this an integration or try shippable build?)

These questions, as you wrote them, seem like they work. `Is this build/push suitable to ship to our users` might be more accurately framed as `is this build configured like a build we would ship to users`, but that may raise more questions. We can clarify further if needed.

rfcs/0005-shippable-builds.md
```diff
@@ -0,0 +1,47 @@
...
+## Motivation
+
+We have a large array of build and test types, as well multiple variants of those types. This goal is to dedupe a
+bunch of very-similar variations into a smaller set that answer the questions we care most about. Which should reduce costs
+of running multiple things that serve similar purposes, and increase confidence in the builds that we do ship.
+
+This work will, in addition, simplify the list of platform variants Sheriffs need to be aware of when triaging checkins, and
+reduce complexity in the taskgraph, enabling us to iterate faster on other future improvements.
+
+There will be a second stage to this work that entails what we call "Nightly Promotion" (RFC to come)
```
escapewindow: During this phase, we're going to have to adjust the release graphs to point at the shippable builds rather than nightly, or we will probably break the first post-merge betas. We probably also have to adjust the nightly graph to use the shippable builds. I'm ok with splitting the work into two if preferred.

In this phase we'd get build consolidation and deal with the renaming. In the nightly promotion phase we'd get convergence of automation and we may regain the usage of proper `run-on-projects` usage for release tasks.

rfcs/0005-shippable-builds.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Details
+
+* We will create a new build variant `shippable` which will replace the variants of 'pgo' and 'nightly'.
+* All tests currently running against either 'pgo' or 'nightly' build types will run against the new 'shippable' variant,
+  and maintain current coverage.
+* We will disable 'pgo' and 'nightly' builds and tests in favor of the new variant.
+* Optimizations, such as SETA or SCHEDULES over when tests will run will remain unchanged.
+* Shippable Builds are always intended to be able to ship to users, so we would have (where practical) zero configuration abnormalities
+  compared to the builds we would want an end-user to have.
```
escapewindow: I'm not 100% sure if the shippable build variants on try and integration will have, say, their update channel configured so they can ship to end-users, but for the most part this is true :) This may fall under the `where practical` clause.

rfcs/0005-shippable-builds.md
tomprince: https://phabricator.services.mozilla.com/D13151 could provide a reasonable stop gap to using shippablbe builds in nightlies.

rfcs/0005-shippable-builds.md
mitchhentges: I'll update mine to follow the directions more closely, good eye, thanks Aki :)

lundjordan:
@Callek - is there anything outstanding for shippable here or can we close this out and add it to the repo given landing of shippable is happening before the EOM?

escapewindow:
> @Callek - is there anything outstanding for shippable here or can we close this out and add it to the repo given landing of shippable is happening before the EOM?

+1, since shippable has landed, we may want to update the PR to reflect reality, point at the resolved bugs, etc., and land this PR.

Callek:
@escapewindow - punting this to you to finish up in any way necessary

escapewindow:
Updated the links to point at the finished bugs. Merging.

PR #7, opened 2019-01-17
Adds docs-location RFC

mitchhentges:
[rendered](https://github.com/mitchhentges/releng-rfcs/blob/docs-location/rfcs/0007-docs-location.md)

### DIFF
--- a/rfcs/0007-docs-location.md
+++ b/rfcs/0007-docs-location.md
@@ -28,18 +28,10 @@ since it's deprecated. Once the last consumer (Thunderbird) moves to ship-it v2,
     additionally, [`scriptworker` docs](https://github.com/mozilla-releng/scriptworker/tree/master/docs) are already a combination of both
     `.rst` and `.md`
 
-When deciding where to add new documentation, the following information can be helpful:
-* The documentation target project's location - if the project is in-tree, perhaps the docs should be as well
-* Do docs for a related project/workflow already exist? If so, perhaps the new docs should be close by
+Additionally, when adding new documentation, use the following tips to decide where they should be placed: 
+* Is the target project for the documentation in-tree? If so, perhaps the docs should be in-tree as well
+* Do docs for a related project/workflow already exist? If so, perhaps the new docs should be written in the same location
 
 # Open Questions
-
-* Should we move all out-of-tree docs to be under a single domain? We could have our 
-top-level `docs.mozilla-releng.net`, plus our projects underneath it: `scriptworker.docs.mozilla-releng.net`, 
-`mozapkpublisher.docs.mozilla-releng.net`, etc
-    * My personal preference is that this won't add much value to the alternative (leaving our other projects under 
-    `readthedocs.io` or other sites), since we'll already be directly linking from `mozilla/build-relengdocs`.
-    The primary benefit that having a single domain will do is avoid dead-link issues if `readthedocs` (or other docs-hosting
-    service that we use) closes down.
 
 # Implementation
--- a/rfcs/0007-docs-location.md
+++ b/rfcs/0007-docs-location.md
@@ -35,10 +35,11 @@ When deciding where to add new documentation, the following information can be h
 # Open Questions
 
 * Should we move all out-of-tree docs to be under a single domain? We could have our 
-top-level `releng.mozilla.org`, plus our projects underneath it: `scriptworker.releng.mozilla.org`, 
-`mozapkpublisher.releng.mozilla.org`, etc
+top-level `docs.mozilla-releng.net`, plus our projects underneath it: `scriptworker.docs.mozilla-releng.net`, 
+`mozapkpublisher.docs.mozilla-releng.net`, etc
+    * My personal preference is that this won't add much value to the alternative (leaving our other projects under 
+    `readthedocs.io` or other sites), since we'll already be directly linking from `mozilla/build-relengdocs`.
+    The primary benefit that having a single domain will do is avoid dead-link issues if `readthedocs` (or other docs-hosting
+    service that we use) closes down.
 
 # Implementation
-
-
-
--- a/rfcs/0007-docs-location.md
+++ b/rfcs/0007-docs-location.md
@@ -22,8 +22,8 @@ Also, update wiki.mozilla.org to describe a summary of the Releng team and link 
 
 ## Motivation
 
-This RFC aims to directly answer the question "where should I put my docs for <x>" or "where are the existing docs 
-for <y>".
+This RFC aims to directly answer the question "where should I put my docs for `$productA`" or "where are the existing docs 
+for `$productB`".
 
 Additionally, this RFC tries to balance the benefit of having centralized documentation with the ability to have docs
 on the same platform as the project they're documenting.
@@ -44,9 +44,9 @@ and [`mozilla/build-relengdocs`](https://github.com/mozilla/build-relengdocs).
 * Comb through `mozilla/build-relengdocs` and remove obsolete documentation
 * Add ability to write docs in `mozilla/build-relengdocs` in markdown
     * Motivation: these docs are on Github and using the Github workflow, and markdown is the common markup language on the platform
-* Move [releaseduty docs](https://github.com/mozilla-releng/releasewarrior-2.0/tree/master/docs) to `mozilla/build-relengdocs`
-* Leave releasewarrior documentation, since it's deprecated. Once the last consumer (Thunderbird) moves to ship-it v2,
-the releasewarrior docs can be removed.
+* Move releaseduty docs ([in here](https://github.com/mozilla-releng/releasewarrior-2.0/tree/master/docs)) to `mozilla/build-relengdocs`
+* Leave releasewarrior documentation ([also in here]((https://github.com/mozilla-releng/releasewarrior-2.0/tree/master/docs))),
+since it's deprecated. Once the last consumer (Thunderbird) moves to ship-it v2, the releasewarrior docs can be removed.
 
 # Open Questions
 
--- a/rfcs/0007-docs-location.md
+++ b/rfcs/0007-docs-location.md
@@ -1,5 +1,5 @@
-# RFC <number> - Location of docs
-* Comments: [#<number>](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/<number>)
+# RFC 7 - Location of docs
+* Comments: [#7](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/7)
 * Proposed by: @mitchhentges
 
 # Summary
--- a/rfcs/0007-docs-location.md
+++ b/rfcs/0007-docs-location.md
@@ -35,8 +35,8 @@ When deciding where to add new documentation, the following information can be h
 * Are we happy with being able to use markdown for `build-relengdocs`?
     * Should we convert existing `.rst` docs in `build-relengdocs` to markdown?
 * Should we move all out-of-tree docs to be under a single domain? We could have our 
-top-level `releng.mozilla.org`, plus our projects underneath it: `scriptworker.mozilla.org`, 
-`mozapkpublisher.mozilla.org`, etc
+top-level `releng.mozilla.org`, plus our projects underneath it: `scriptworker.releng.mozilla.org`, 
+`mozapkpublisher.releng.mozilla.org`, etc
 
 # Implementation
 
--- a/rfcs/0007-docs-location.md
+++ b/rfcs/0007-docs-location.md
@@ -23,8 +23,10 @@ Additionally, there's some minor fragmentation and documentation accessibility t
 * Update https://wiki.mozilla.org/Main_Page - move technical information to `mozilla/build-relengdocs`
     * Note: leave releasewarrior documentation ([also in here]((https://github.com/mozilla-releng/releasewarrior-2.0/tree/master/docs))),
 since it's deprecated. Once the last consumer (Thunderbird) moves to ship-it v2, these docs can be removed.
-* Add ability to write docs in `mozilla/build-relengdocs` in markdown
-    * Motivation: these docs are on Github and using the Github workflow, and markdown is the common markup language on the platform 
+* Add ability to write docs in `mozilla/build-relengdocs` in markdown (while still allowing `.rst`)
+    * Motivation: these docs are on Github and using the Github workflow, and markdown is the common markup language on the platform.
+    additionally, [`scriptworker` docs](https://github.com/mozilla-releng/scriptworker/tree/master/docs) are already a combination of both
+    `.rst` and `.md`
 
 When deciding where to add new documentation, the following information can be helpful:
 * The documentation target project's location - if the project is in-tree, perhaps the docs should be as well
@@ -32,8 +34,6 @@ When deciding where to add new documentation, the following information can be h
 
 # Open Questions
 
-* Are we happy with being able to use markdown for `build-relengdocs`?
-    * Should we convert existing `.rst` docs in `build-relengdocs` to markdown?
 * Should we move all out-of-tree docs to be under a single domain? We could have our 
 top-level `releng.mozilla.org`, plus our projects underneath it: `scriptworker.releng.mozilla.org`, 
 `mozapkpublisher.releng.mozilla.org`, etc
--- a/rfcs/0007-docs-location.md
+++ b/rfcs/0007-docs-location.md
@@ -4,54 +4,39 @@
 
 # Summary
 
-When adding new documentation:
-* If it's for an in-tree project:
-    1. Add to [in-tree docs](https://firefox-source-docs.mozilla.org/taskcluster/taskcluster/release-promotion.html)
-    2. Add a link to the in-tree docs in [`mozilla/build-relengdocs`](https://github.com/mozilla/build-relengdocs) 
-* Otherwise:
-    1. Add to [`mozilla/build-relengdocs`](https://github.com/mozilla/build-relengdocs)
-    
-When looking for documentation for an existing project:
-1. Go to [`mozilla/build-relengdocs`](https://github.com/mozilla/build-relengdocs)
-2. If it links to in-tree docs, follow the link
-
-As part of this RFC, we'll coalesce existing documentation into the in-tree and `mozilla/build-relengdocs` docs so
-they match the decision tree above.
-
-Also, update wiki.mozilla.org to describe a summary of the Releng team and link to `mozilla/build-relengdocs`
+For all of our releng documentation, ensure that it's possible to find via `mozilla/build-relengdocs`.
+This may be accomplished for each piece of documentation by either:
+* Storing it in `mozilla/build-relengdocs`
+* Storing it elsewhere, but linking to the published location from `mozilla/build-relengdocs`
 
 ## Motivation
 
-This RFC aims to directly answer the question "where should I put my docs for `$productA`" or "where are the existing docs 
-for `$productB`".
+This RFC aims to ease the task of answering "where are the existing docs for `$product` or `$workflow`" by ensuring that
+docs are findable from a central location (`mozilla/build-relengdocs`).
 
-Additionally, this RFC tries to balance the benefit of having centralized documentation with the ability to have docs
-on the same platform as the project they're documenting.
-
-Requirements for documentation platforms:
-* Smaller changes (e.g.: typo fixes) are possible without waiting for review
-* Larger changes can be reviewed with a nice diff UI (GitHub's does this nicely)
-* We can view history of docs
-* The published docs are easily accessible (e.g.: $something.readthedocs.io)
-* Process of updating docs shouldn't be too heavy: making it tough to edit docs may increase the chances that docs will become out-of-date
+Additionally, there's some minor fragmentation and documentation accessibility that can be improved.
 
 # Details
 
-Have two platforms for documentation: [in-tree docs](https://firefox-source-docs.mozilla.org/taskcluster/taskcluster/release-promotion.html)
-and [`mozilla/build-relengdocs`](https://github.com/mozilla/build-relengdocs).
-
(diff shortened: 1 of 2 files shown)


rfcs/0007-docs-location.md
```diff
@@ -0,0 +1,36 @@
+# RFC 7 - Location of docs
+* Comments: [#7](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/7)
+* Proposed by: @mitchhentges
+
+# Summary
+
+* Update wiki.mozilla.org to describe a summary of the Releng team, links to more-detailed docs
+* Remove releng.readthedocs.io from gecko in-tree
```
tomprince: I'm not sure these live in mozilla-central. I think they may come from https://github.com/mozilla/build-relengdocs (via following the "view on github link on https://moz-releng-docs.readthedocs.io/en/latest/)

rfcs/0007-docs-location.md
```diff
@@ -0,0 +1,36 @@
...
+* Create a new repository similar to [`taskcluster-docs`](https://github.com/taskcluster/taskcluster-docs). Move information from releng.readthedocs.io to here
+
+
+## Motivation
+
+We need a place to house docs that meets the following requirements:
+
+* Smaller changes (e.g.: typo fixes) are possible without waiting for review
```
tomprince: This is definitely possible in mozilla-central/hg (though as noted above the docs you are pointing to aren't there).

rfcs/0007-docs-location.md
mitchhentges: Oh, we already have a repository, interesting :thinking: Perhaps we should continue using `mozilla/build-relengdocs`?
I'll add that as a question to the RFC, we didn't discuss this in our mobile releng meeting

rfcs/0007-docs-location.md
```diff
@@ -28,6 +28,9 @@ Follow the approach laid out in [`taskcluster-docs`](https://github.com/taskclus
 * Should we have releng-desktop.readthedocs.io and releng-mobile.readthedocs.io?
     * Should we have two separate docs sites, but tweak the above names?
     * Should we have all releng docs shared in a single site?
+* How should we handle the [in-tree docs](https://firefox-source-docs.mozilla.org/taskcluster/taskcluster/release-promotion.html)
+    * Move them to our new docs repository? Keep the in-tree docs, but have define how to decide whether new docs go in-tree or in-new-repository?
```
tomprince: Given that they are specific to in-tree code, I think keeping them in-tree makes sense.

rfcs/0007-docs-location.md
```diff
@@ -28,6 +28,9 @@ Follow the approach laid out in [`taskcluster-docs`](https://github.com/taskclus
...
+* Should we handle [`warrior-docs`](https://github.com/mozilla-releng/releasewarrior-2.0/tree/master/docs) as part of this RFC, or wait until Ship-it v2 to migrate into that?
```
tomprince: I think moving the releaseduty docs to the new repo makes sense, particularly as releasewarrior as a tool is only used for TB (which is hopefully going to move to ship-it v2 soon).

The old ship-it v1 docs for TB should probably stay until it has migrated.

rfcs/0007-docs-location.md
```diff
@@ -0,0 +1,38 @@
...
+* Update wiki.mozilla.org to describe a summary of the Releng team, links to more-detailed docs
+* Remove [`moz-releng-docs.readthedocs.io`](https://moz-releng-docs.readthedocs.io/en/latest/) from [`mozilla/build-relengdocs`](https://github.com/mozilla/build-relengdocs)
+* Create a new repository similar to [`taskcluster-docs`](https://github.com/taskcluster/taskcluster-docs). Move information from releng.readthedocs.io to here
+
+## Motivation
+
+We need a place to house docs that meets the following requirements:
+
+* Smaller changes (e.g.: typo fixes) are possible without waiting for review
+* Larger changes can be reviewed with a nice diff UI (GitHub's does this nicely)
```
tomprince: Personally, I'd prefer hg and phabricator, but I suspect I'm in the minority.

mitchhentges: I'm less familiar with `hg` and `phabricator`, and I think that there's benefit to be gained by leveraging how prevalent `git` and `GitHub`'s tooling are.

However, this argument ("use this because it's popular, new hires will already know how to use it!") is the same argument that puts companies on the JS framework treadmill, so it's not as strong of an argument as it seems at face value.

rfcs/0007-docs-location.md
tomprince: > We know that docs are specific to in-tree because they're Firefox-desktop-specific, is that correct?
Firefox and Fennec specific, both of which live in-tree.

> Is the `gecko` engine that we use for mobile also in-tree?

Yes.

rfcs/0007-docs-location.md
```diff
@@ -0,0 +1,38 @@
...
+* The published docs are easily accessible (e.g.: $something.readthedocs.io)
+* `moz-releng-docs.readthedocs.io` hasn't been updated for a few years, if we want to change the location of our docs, we wouldn't be harming our existing workflow too harshly
+
+# Details
+
+Follow the approach laid out in [`taskcluster-docs`](https://github.com/taskcluster/taskcluster-docs). Deploy these docs to a `readthedocs` site or equivalent.
+
+# Open Questions
+
+* Instead, should we continue using `mozilla/build-relengdocs`?
```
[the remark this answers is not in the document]
catlee: What advantages are there for creating a new repository and deprecating the old one? IMO, if we already have a docs repo, let's make it work for us instead of creating new one and adding yet another source of documentation.

rfcs/0007-docs-location.md
```diff
@@ -0,0 +1,38 @@
...
+* Should we have releng-desktop.readthedocs.io and releng-mobile.readthedocs.io?
+    * Should we have two separate docs sites, but tweak the above names?
+    * Should we have all releng docs shared in a single site?
```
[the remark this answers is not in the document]
mitchhentges: Hmm, we have some mutually exclusive options here:
* As [Tom noted](https://github.com/mozilla-releng/releng-rfcs/pull/7/files#r249156750), keeping docs specific to in-tree code in-tree keeps 'em closer to the projects they document
* As you've noted here, some projects span multiple systems. Arguably, if we kept docs beside the projects they document, we wouldn't need to worry about the location of the the systems that use them. For example, for geckoview, since it's in-tree, the docs would be in-tree as well (even though depend systems are on Github)
* The downside to having docs be beside their projects is that docs become fragmented (where do I find docs for <x>?) Is this a greater or lesser cost than having to update the monolithic docs location that has little relation to the project you're working on?
* FWIW, I'm very used to workflows with Github. How challenging would it be to set up Github-stockholm-sydrome'd devs like myself with `hg` and doing docs in-tree? @tomprince maybe we can do a sweet vidyo/IRC Q&A so I can see the glorious non-Github workflow?

rfcs/0007-docs-location.md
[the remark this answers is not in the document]
mitchhentges: True, though we'd need to spend some time pruning through those older docs to find what hasn't succumbed to bitrot, ensuring deployment is still functional, etc :)
Though, migration will make sure we don't lose anything and keeps the domain compatible. If we decide on having a Github docs repository, I'd prefer to keep `mozilla/build-relengdocs` rather than creating a new repo

Note that this is dependent a bit on the conversation [here](https://github.com/mozilla-releng/releng-rfcs/pull/7/files#r253689089) - if we decide to keep all of our docs together in the same repo, then _some_ docs are going to have to move (either this repo or in-tree's).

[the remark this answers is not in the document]
escapewindow: I'd be ok with having this repo be the overall source of docs, which contains links to the in-tree gecko docs where applicable.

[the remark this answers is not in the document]
mitchhentges: How would you define "overall source of docs"?
I'd like to be able to define how to decide where future, new documentation should go.
We could do: "when documenting a project that exists in-tree, then add to in-tree docs. Otherwise, add to `mozilla/build-relengdocs`." Would that be :+1:?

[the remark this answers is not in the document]
escapewindow: When documentint a project that exists in-tree, then add to in-tree docs, and link to it from build-relengdocs. That way you could find all docs from a single location, though some would be linked.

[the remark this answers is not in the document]
mitchhentges: That sounds fantastic :+1: Thanks!

rfcs/0007-docs-location.md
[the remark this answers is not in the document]
mitchhentges: Moving this conversation to the other thread and updating the RFC with the points Aki made

mitchhentges:
@tomprince @MihaiTabara @escapewindow I've attempted to formalize your notes in my recent updates to the RFC, would you mind giving it another look?
[Rendered](https://github.com/mitchhentges/releng-rfcs/blob/docs-location/rfcs/0007-docs-location.md)

rfcs/0007-docs-location.md
```diff
@@ -0,0 +1,59 @@
+# RFC 7 - Location of docs
+* Comments: [#7](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/7)
+* Proposed by: @mitchhentges
+
+# Summary
+
+When adding new documentation:
+* If it's for an in-tree project:
+    1. Add to [in-tree docs](https://firefox-source-docs.mozilla.org/taskcluster/taskcluster/release-promotion.html)
+    2. Add a link to the in-tree docs in [`mozilla/build-relengdocs`](https://github.com/mozilla/build-relengdocs)
```
escapewindow: Hm, we might want to add another section for things like scriptworker, which have their [own docs](https://scriptworker.readthedocs.io/en/latest/): similar to in-tree, but for standalone projects.

mitchhentges: Ahh, great point! I forgot about `scriptworker` having its own site rn
Do we want `scriptworker` to have its own independent docs site? IMHO, there's benefit for in-tree docs to be separate - editing in-tree docs is the same workflow as in-tree projects.
Meanwhile, for `scriptworker`, it's already the same workflow and platform as `mozilla/build-relengdocs`. I'm leaning towards moving `scriptworker` docs to be part of `mozilla/build-relengdocs`

escapewindow: I think there is value: it's a worker for taskcluster that other teams outside of releng use. It's easier to find the docs for it through the repo, rather than pointing them at a section in our releng docs. There may be other projects with their own docs sites; as long as they're discoverable from a single source (which we solve here via linking), I'm good.

escapewindow: Similarly, for docs that we don't want to be public, we may link to the mana.

mitchhentges: >  It's easier to find the docs for it through the repo, rather than pointing them at a section in our releng docs.

Either way, we need to link from the repository to rendered docs, right? (I'm assuming that consumers would prefer rendered docs than the source markdown/`.rst` material). Once it's just a link, it doesn't matter to the reader if it points to a standalone page, or a subsection in a larger site, right?

If `scriptworker` has its own documentation site, then I fear that we're going to slippery-slope into a state where we have many different locations for docs.

-----

Related: how do we want to handle generated docs (pydoc)?

So far, I've just been considering documentation for workflows - custom guides and so on. These are easily moved to new locations, since they're relatively standalone documentation.

However, generated docs are always sourced from the code itself.

One solution we can opt for (temporarily ignoring in-tree, since I'm less familiar) is:
* Move all workflow docs (scriptworker, releaseduty, etc) to `mozilla/build-relengdocs`. Host these docs at https://moz-releng-docs.readthedocs.io/ (maybe even our own subdomain?)
* For generated docs, have them at a subdomain. For example, `scriptworker` would be at https://scriptworker.moz-releng-docs.readthedocs.io/
* In each project readme, link to both workflow and generated docs

What do you think? Was it these pydocs that you were referring to in your last comment?

escapewindow: That could work. Process docs wouldn't be able to ride a PR, but we can have "as of VERSION" or something in the docs.

escapewindow: And I was referring to both, but yes, the pydocs are easier to keep close to the code.

mitchhentges: I'm assuming that a significant amount of process docs will touch more than one project, so we're going to have the problem of manually synchronizing those docs with their project versions regardless of our solution

rfcs/0007-docs-location.md
mitchhentges: Regarding in-tree docs: are you referring to specifically generated (e.g.: `pydoc`) docs?
I'm :100:% on the same page that generated docs _need_ to stay in-tree, since they're beside the code that they document.

However, are there any in-tree docs about workflows? What do you think about moving all workflow documentation to a single location, like that `mozilla/build-relengdocs` repo and site?

escapewindow: I'm probably -1 on that. The taskgraph docs for gecko are all in the gecko in-tree docs, including workflow.

mitchhentges: Ok, gotcha.
I'm a little confused - what would be the potential downside of moving workflow taskgraph/gecko docs from in-tree to a separate repository?

The benefit I'm looking at is: maybe we can centralize all workflow docs in a single location, but leave generated docs in their associated locations.

is the downside that you're seeing the fact that people are used to finding them in their current location? Or are the workflow docs hard to physically remove from in-tree? Thanks :)

escapewindow: Someone hacking on the in-tree taskgraph code will likely need information from both releng-specific and non-releng-specific docs, and they're already in a single location. Splitting them breaks rtd search, easy single-location reading, and an easy way to maintain a constant style. It also makes it harder to keep the docs riding the trains, which happens regularly with the in-tree docs.

mitchhentges: Hmm, how can we define a deterministic way of figuring out where docs should go? Sounds like the "all workflow docs in `mozilla/build-relengdocs`" strategy is a no-go :)

How about the following:
* All generated documentation will still remain beside source code.
    * For in-tree docs, they remain deployed to the same location
    * For github-based docs, their deployment location will change to `https://$product.moz-releng-docs.readthedocs.io`
* All workflow docs specific to releng (not used by other teams) should be in `mozilla/build-relengdocs`
* Other workflow docs that span between multiple teams remain in their current location. The only instance of this case (that I know of so far) is the in-tree docs

escapewindow: > Hmm, how can we define a deterministic way of figuring out where docs should go? Sounds like the "all workflow docs in `mozilla/build-relengdocs`" strategy is a no-go :)

I think a rule of thumb is fine.

>       * For in-tree docs, they remain deployed to the same location
>       * For github-based docs, their deployment location will change to `https://$product.moz-releng-docs.readthedocs.io`

Is this something you know works with rtd? I haven't done it before, but that doesn't mean it isn't possible :)

There are [subprojects](https://docs.readthedocs.io/en/latest/subprojects.html). Also [this](https://docs.readthedocs.io/en/latest/faq.html#how-do-i-change-my-project-slug-the-url-your-docs-are-served-at) and [this](https://docs.readthedocs.io/en/latest/canonical.html). We may have some figuring out to do if we want the webhooks to fire across multiple github repos; if they're repo-specific that should work out of the box.

>     * All workflow docs specific to releng (not used by other teams) should be in `mozilla/build-relengdocs`

I think a rule of thumb is sufficient, as long as things are linked in the releng docs. The second main bullet point sums it up.

>     * Other workflow docs that span between multiple teams remain in their current location. The only instance of this case (that I know of so far) is the in-tree docs

Scriptworker does span multiple teams, but is small enough that it could have split docs.
Could go either way.

tomprince: > > how can we define a deterministic way of figuring out where docs should go?
> I think a rule of thumb is fine.
I strongly agree that having heuristics is enough.

mitchhentges: My concern with having a heuristic is that it's tough to find docs - it becomes a trial-and-error game of going to the `$n` different places where relevant docs might be, e.g.:
* In-tree's site?
* In `mozilla/build-relengdocs`'s site?
* Releasewarrior's old site?
* Something else altogether?

-----

I've been trying to somehow centralize our docs according to [this comment by Johan](https://github.com/mozilla-releng/releng-rfcs/pull/7#discussion_r252737444), because of my current understanding:
* If we resort to a rule of thumb for deciding where documentation goes, then that translates to mostly "staying the course" for how documentation is accessed right now
* We aren't happy with the current state of documentation since it's fragmented across multiple locations

It sounds like I'm misunderstanding something :thinking:

tomprince: I think we can have central place to *find* documentation, without all the documentation living there. So, wherever documents live, there will be a pointer from a single location, that points to all the place documentation might live.

mitchhentges: FWIW, would there be a benefit to that approach rather than linking from code repository to docs?
So, from the readers perspective, instead of:
1. Go to centralized location
2. Follow link to relevant docs

They would instead:
1. Find project in-tree/in-github
2. Follow link to docs

With this approach, we wouldn't need to keep that centralized location up-to-date. Additionally, we wouldn't need to worry about the granularity of links from the centralized location. For example, should we link for every workflow guide (oof) or on a project-wide basis? If on a project-wide basis, what if docs for that product exist in multiple locations?

On the other hand, with this lightweight approach, it's not obvious where docs will be if you don't know where the project lives (is geckoview GH or in-tree? Is scriptworker GH or in-tree? etc).

Throwing this out as a potential related solution. What do you think?

escapewindow: I think it should be linked from both locations, because we will likely have people looking for docs from both locations:

- I'm reading releng docs, and want to know about something in-tree: I find the docs link in the releng docs rtd
- I'm looking at a code repo and want to read docs about it: I find the docs link in the code repo's readme

tomprince: One benefit of having docs next to the code, is that they can be changed in lock-step, as part of the commit. It is even possible (m-c's taskgraph code does this) to have tests that things are documented.

mitchhentges: > One benefit of having docs next to the code, is that they can be changed in lock-step, as part of the commit

This is really true, though it's less applicable when a piece of documentation applies to multiple projects that aren't in the same repository, and therefore can't be changed in lock-step

-----

I'll update the RFC to increase flexibility, thanks for the insight! :fire: :ok_hand:

rfcs/0007-docs-location.md
```diff
@@ -0,0 +1,44 @@
...
+    * Motivation: these docs are on Github and using the Github workflow, and markdown is the common markup language on the platform 
+
+When deciding where to add new documentation, the following information can be helpful:
+* The documentation target project's location - if the project is in-tree, perhaps the docs should be as well
+* Do docs for a related project/workflow already exist? If so, perhaps the new docs should be close by
+
+# Open Questions
+
+* Are we happy with being able to use markdown for `build-relengdocs`?
+    * Should we convert existing `.rst` docs in `build-relengdocs` to markdown?
```
mitchhentges: If we're embracing the Github ecosystem, markdown is the preferred flavour for documentation. Additionally, (IMHO), markdown is lighter-weight to write by hand (e.g.: in `.rst`, titles need a line before and after each title filled with `===`, and these lines need to be the same length as the title. When the title is changed, formatting busywork is required).

FWIW (if we agree that markdown is more acceptable), the busywork cost of refactoring from `.rst` to `.md` is pretty automate-able with `pandoc`

rfcs/0007-docs-location.md
```diff
@@ -0,0 +1,44 @@
...
+* Should we move all out-of-tree docs to be under a single domain? We could have our
```
mitchhentges: Is this worth pursuing? Since we work on some tooling that isn't releng-specific, I'm not sure that this makes sense
@lundjordan do we own a releng domain?

rfcs/0007-docs-location.md
escapewindow: I agree that for the most part, `.md` is easier to deal with. However, if you want to create more complex documentation, `.md` can't handle everything that `.rst` can, so `.rst` is the better choice there. And then it's a matter of whether consistency across the whole source is more important, or if ease of editing the less-complex docs is more important.

mitchhentges: IMHO, consistency across the whole source is more important.

> `.md` can't handle everything that `.rst` can

Do you have any `.rst` examples handy that we use?

escapewindow: Our [in-tree docs](https://hg.mozilla.org/mozilla-central/file/tip/taskcluster/docs) are rst. [scriptworker docs](https://github.com/mozilla-releng/scriptworker/tree/master/docs) are a mix, depending on whether we need added functionality. I think one issue I hit is you can't include a markdown file inside of another one. Searching around, toctree may be easier with rst than md? [Here](https://github.com/JuliaFEM/JuliaFEM.jl/issues/49#issue-93635416) are some arguments for both.

If we only want one format, we probably should go with the more fully featured format. If we're open to both formats, we can choose the one that best fits.

I think trying to make a strict set of rules here is likely going to be difficult. As I alluded to earlier, I think it's best to have a location for all docs to live in or be linked from, and have some rules of thumb for what we prefer. Even if we don't have everyone's preference, the discoverability alone should be a win.

mitchhentges: Sounds great, let's keep it more flexible :)

lundjordan:
are we at stalemate? Is there some imperfect solution we could go with? e.g. centralize some docs and link to others from that centralized location?

escapewindow:
+1 to centralize with links.

mitchhentges:
Bumping to "Final Comment" since there's no open questions and this has settled down

escapewindow:
Let's land this!

PR #8, opened 2019-01-17
0008 - scriptworker *script monorepo

escapewindow:
We should put all of our releng-maintained production scriptworker *scripts into a single repo, so we can improve discoverability, keep \*script behaviors consistent, and make it easier to make *script-wide changes.

### DIFF
--- a/rfcs/0008-scriptworker-script-monorepo.md
+++ b/rfcs/0008-scriptworker-script-monorepo.md
@@ -46,8 +46,10 @@ We need to have ways of excluding or special-casing \*scripts. `balrogscript` is
 
 We should version each module or script like we do currently: semver. Libraries (e.g., scriptworker-client) will bump major versions every time they make a non-backwards-compatible change. The \*scripts will bump their major versions when they make a task- or puppet- (runtime- or deployment-) backwards-incompatible change.
 
+Initially, we will push docker releases via pushes to production branches. In the future, we will port scriptworker-scripts to taskgraph, and take advantages of the features it has (only build what changes; Chain of Trust verification; using artifacts and indexes from previously run tasks).
+
 # Open Questions
 
 # Implementation
 
-I'm playing with an experimental monorepo at [escapewindow/scriptworker-scripts](https://github.com/escapewindow/scriptworker-scripts). We can go with this, make adjustments, or completely revisit, depending on what people think.
+We have a monorepo at [mozilla-releng/scriptworker-scripts](https://github.com/mozilla-releng/scriptworker-scripts), already using Taskcluster for automation. The scripts are yet to be unified in terms of behavior, using `scriptworker_client`, using taskgraph, etc., but we will do that over time.
--- a/rfcs/0008-scriptworker-script-monorepo.md
+++ b/rfcs/0008-scriptworker-script-monorepo.md
@@ -32,6 +32,7 @@ We are still able to make exceptions for behavior in \*scripts when needed. For 
 
 # Details
 
+## Proposal 1
 We create a github repo, `mozilla-releng/scriptworker-scripts`. We move the code from `scriptworker.client` into a `scriptworker-client` subdirectory, with its own `setup.py`. Tests will go under each subdirectly, e.g. `scriptworker-client/tests/` or `signingscript/tests`. We can move each production-ready, releng-maintained \*script into the monorepo over time.
 
 Each \*script will have its own `setup.py`.
@@ -44,6 +45,19 @@ We can do things like pyup dependency pinning across the repo like we do with pu
 
 We need to have ways of excluding or special-casing \*scripts. `balrogscript` is py2; the upcoming `applescript` will be targeted towards running on mac hosts instead of deployed through kubernetes.
 
+## Proposal 2
+We gradually begin moving our \*scripts into the [release-services](https://github.com/mozilla/release-services) repo, as we nixify and prepare each one for deployment into GCP.
+
+Each \*script will still have its own `setup.py` and its own `tests/` subdirectory.
+
+Any change in shared behavior should still span all \*scripts as applicable.
+
+We need to adjust the release-services taskgraph and deployment story to also support \*script deployment processes and cadences. Most likely we need a way to easily release a subset of projects inside the repo, and exclude projects from the current services release cadence.
+
+Most likely our pinning will happen through nix.
+
+We need to have ways of excluding or special-casing \*scripts. `balrogscript` is py2; the upcoming `applescript` will be targeted towards running on mac hosts instead of deployed through kubernetes.
+
 # Open Questions
 
 We need a way to handle applescript and balrogscript differently from the rest of the \*scripts.
@@ -52,7 +66,7 @@ Do we call it mozilla-releng/scriptworker-scripts ?
 
 Specific details about directory layouts, tox.ini per subdirectory or top level?, pyup pinning standards? nix?, code coverage service?, but I imagine we will decide on something good, and have the capability of changing it in the future without having to touch 10+ repos for consistency.
 
-Do we combine all releng [python] repos into one?
+Do we go with option 1 or 2? (\*scripts living with \*services or no?)
 
 Versioning and tagging rules. Does Nix make this obsolete?
 
--- a/rfcs/0008-scriptworker-script-monorepo.md
+++ b/rfcs/0008-scriptworker-script-monorepo.md
@@ -48,11 +48,13 @@ We need to have ways of excluding or special-casing \*scripts. `balrogscript` is
 
 We need a way to handle applescript and balrogscript differently from the rest of the \*scripts.
 
-We haven't touched binary transparency in a long time; perhaps we should call this not-yet-production-ready and wait to move it?
-
 Do we call it mozilla-releng/scriptworker-scripts ?
 
 Specific details about directory layouts, tox.ini per subdirectory or top level?, pyup pinning standards? nix?, code coverage service?, but I imagine we will decide on something good, and have the capability of changing it in the future without having to touch 10+ repos for consistency.
+
+Do we combine all releng [python] repos into one?
+
+Versioning and tagging rules. Does Nix make this obsolete?
 
 # Implementation
 
--- a/rfcs/0008-scriptworker-script-monorepo.md
+++ b/rfcs/0008-scriptworker-script-monorepo.md
@@ -50,3 +50,4 @@ We should version each module or script like we do currently: semver. Libraries 
 
 # Implementation
 
+I'm playing with an experimental monorepo at [escapewindow/scriptworker-scripts](https://github.com/escapewindow/scriptworker-scripts). We can go with this, make adjustments, or completely revisit, depending on what people think.
--- a/rfcs/0008-scriptworker-script-monorepo.md
+++ b/rfcs/0008-scriptworker-script-monorepo.md
@@ -45,6 +45,16 @@ We can do things like pyup dependency pinning across the repo like we do with pu
 
 We need to have ways of excluding or special-casing \*scripts. `balrogscript` is py2; the upcoming `applescript` will be targeted towards running on mac hosts instead of deployed through kubernetes.
 
+Pluses:
+
+- the \*scripts have a lot in common, and will behave the same.
+- we can potentially create this monorepo sooner, and adjust behavior as we nixify each \*script and deploy to GCP instead of through puppet
+
+Concerns:
+
+- we'll have shared code+automation needs between release-services and \*scripts, e.g. papertrail logging and nix support.
+- we may end up moving the \*scripts a second time if it makes sense to share a repo with release-services in the future.
+
 ## Proposal 2
 We gradually begin moving our \*scripts into the [release-services](https://github.com/mozilla/release-services) repo, as we nixify and prepare each one for deployment into GCP.
 
@@ -57,6 +67,17 @@ We need to adjust the release-services taskgraph and deployment story to also su
 Most likely our pinning will happen through nix.
 
 We need to have ways of excluding or special-casing \*scripts. `balrogscript` is py2; the upcoming `applescript` will be targeted towards running on mac hosts instead of deployed through kubernetes.
+
+Pluses:
+
+- we can share code and automation between release-services and \*scripts, which will look more and more similar when we move to GCP
+- one place to make improvements across all of our microservices
+- a larger pool of people familiar with each workflow. `bus_factor++`
+
+Concerns:
+- we probably don't want to move the \*scripts in until they're ready for nix+docker+GCP, making this a longer term project
+- we'll need to adjust the release-services shared code, automation, and cadences to allow for similar-but-different \*script code+automation+cadence needs.
+- the `please` tool has some rough edges, which we'll want to address
 
 # Open Questions
 
--- a/rfcs/0008-scriptworker-script-monorepo.md
+++ b/rfcs/0008-scriptworker-script-monorepo.md
@@ -13,18 +13,18 @@ We have 10 \*scripts in puppet currently, and are looking at adding more. Curren
 This proves challenging for:
 
 * discoverability. If you want to find all of the scriptworker \*scripts, you can browse the mozilla-releng organization for any repos that seem to match the right naming scheme. As the number of repos increase, or if a \*script is outside the mozilla-releng org for any reason, it will get worse.
-* consistency in \*script behavior.
-* consistency in github automation.
-* consistency in \*script deployment.
-* making changes across \*scripts.
+* consistency in \*script behavior. Outside of recommended behaviors and the shared code in `scriptworker.client`, the separate repos lend themselves to divergence of behavior.
+* consistency in github automation. Similar to the \*script behavior, the travis+pyup+code coverage automation configurations have already diverged.
+* consistency in \*script deployment. With each maintainer defining their own behaviors, we've deployed some \*scripts to pypi, added towncrier to some.
+* making changes across \*scripts. We've had a few instances of this in recent memory, where someone has to create PRs across 10+ repos.
 
 Moving to a monorepo for \*scripts improves all of the above:
 
 * discoverability: if you find the monorepo, you've found all of the \*scripts.
 * consistency in \*script behavior: it's possible to inspect and modify \*scripts' behavior in a single place.
-* consistency in github automation:
-* consistency in \*script deployment:
-* making changes across \*scripts:
+* consistency in github automation: with a single `.travis.yml` and/or `.taskcluster.yml`, and a single repo to point other services at (e.g. code coverage, pyup).
+* consistency in \*script deployment: we're currently planning on converging on nix, but even nix-based solutions could diverge over 10+ repos. The monorepo allows us to configure them all from the same place.
+* making changes across \*scripts: even if we want to make a sweeping change, we could find the relevant scripts in the same repo.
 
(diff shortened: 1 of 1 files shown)


rfcs/0008-scriptworker-script-monorepo.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Open Questions
+
+We need a way to handle applescript and balrogscript differently from the rest of the \*scripts.
+
+We haven't touched binary transparency in a long time; perhaps we should call this not-yet-production-ready and wait to move it?
+
+Do we call it mozilla-releng/scriptworker-scripts ?
+
+Specific details about directory layouts, tox.ini per subdirectory or top level?, pyup pinning standards? nix?, code coverage service?, but I imagine we will decide on something good, and have the capability of changing it in the future without having to touch 10+ repos for consistency.
```
escapewindow: We could either implement nix as the way we deploy from the monorepo, and have that be a requirement to land, or we could have a non-nix solution and a nix solution living together with a config file specifying which *scripts use which method.

rfcs/0008-scriptworker-script-monorepo.md
```diff
@@ -0,0 +1,58 @@
...
+
+We can do things like pyup dependency pinning across the repo like we do with puppet.
+
+We need to have ways of excluding or special-casing \*scripts. `balrogscript` is py2; the upcoming `applescript` will be targeted towards running on mac hosts instead of deployed through kubernetes.
+
+# Open Questions
+
+We need a way to handle applescript and balrogscript differently from the rest of the \*scripts.
+
+We haven't touched binary transparency in a long time; perhaps we should call this not-yet-production-ready and wait to move it?
```
tomprince: :+1:

rfcs/0008-scriptworker-script-monorepo.md
```diff
@@ -0,0 +1,58 @@
...
+
+Do we call it mozilla-releng/scriptworker-scripts ?
+
+Specific details about directory layouts, tox.ini per subdirectory or top level?, pyup pinning standards? nix?, code coverage service?, but I imagine we will decide on something good, and have the capability of changing it in the future without having to touch 10+ repos for consistency.
+
```
tomprince: More open questions:
- versioning and tagging
- release process and release notes (towncrier across the board?)

rfcs/0008-scriptworker-script-monorepo.md
garbas: we already turned shipitscript into a nix project and we are deploying it to cloudops soon via release-services setup: https://github.com/mozilla/release-services/pull/1803

now this is just a temporary for now, but i can see us pulling in one scriptworker at the time (as we migrate to cloudops) into shipitscript repo (or whatever we will rename it too) and turning it into a monorepo.

garbas:
> We should put all of our releng-maintained production scriptworker *scripts into a single repo, so we can improve discoverability, keep *script behaviors consistent, and make it easier to make *script-wide changes.

to play a bit of a devils advocate, why are we not moving all the code into one repo and benefit even more from sharing between projects? :smile: how to setup logging should not really be that different from project to project, right?

rfcs/0008-scriptworker-script-monorepo.md
garbas: if nix is used as in release-services repository then we already documented release process: https://docs.mozilla-releng.net/deploy/regular.html

and while you can use version per project nobody is really using it since it became we don't really benefit from it. nix kinda solved this. what i want to say is, you can use whatever system you want in release-services, but nobody is, because of nix. i guess until you start using release-services you wouldn't really know what i mean, but i tried to explain anyway :smile:

escapewindow:
> to play a bit of a devils advocate, why are we not moving all the code into one repo and benefit even more from sharing between projects? 😄 how to setup logging should not really be that different from project to project, right?

If you mean the *scripts should move to release-services, that's a very interesting idea. Since we're planning on using nix to create docker images, that's a very similar workflow to the services docker images. We'd probably want to protect both with things like dockerhub scriptworker, and would have to either haggle about processes and rules, or have separate sets of rules per project in the repo.

If you mean putting scriptworker in with the *scripts, or with release-services, it's a standalone project that I'm trying to create a more defined handoff point (calling the *script with a config file and possibly other commandline options, but not having the *script import scriptworker.) I'm less convinced that it belongs in the same repo.

rfcs/0008-scriptworker-script-monorepo.md
escapewindow: Nice! We should plan on adding chain of trust and non-tcsecrets-deployment to this and other sensitive docker image processes in q2 or q3.

escapewindow:
> > If you mean the *scripts should move to release-services, that's a very interesting idea. Since we're planning on using nix to create docker images, that's a very similar workflow to the services docker images. We'd probably want to protect both with things like dockerhub scriptworker, and would have to either haggle about processes and rules, or have separate sets of rules per project in the repo.
> 
> I'm not sure what you mean with protecting things like dockerhub scriptworker, if you have some time would you care to explain it to me.

Sure, I'm leaving this for the meeting.

> And regarding processes, I don't think scriptworker future processes are (or should be) any different from let say how we deploy shipit. Not this is only my assumption and I might be wrong since I don't see the whole picture. I think we can all see a value of having the same (or similar) process for all projects, if this is possible of course. I'm definitely willing to change how we do things for existing projects in release services to accommodate wider range of projects. I don't think current project in release services are special in any regard.

We can discuss this in the meeting. Essentially it's a worker like generic-worker; other non-releng teams use it. We're able to build it into our docker images like any other upstream dependency, but other teams or projects can use it without going through nix or docker if wanted.

> I think we are now at the point where we would need to meet over vidyo and flesh out details how any of above mentioned options will look like. especially considering time constrains and our other work.

Per the [mechanics](https://github.com/mozilla-releng/releng-rfcs/blob/master/mechanics.md#open-pr-for-comments), we're still in the draft phase. The conversation sounds like a good idea.

> @escapewindow do you want to setup a meeting for this? I can join you once my kinds are in bed (eg. after 20:30CET).

Sure, done.

escapewindow:
This is ready for wider review: @mozilla-releng/releng 

[rendered](https://github.com/escapewindow/releng-rfcs/blob/script-monorepo/rfcs/0008-scriptworker-script-monorepo.md)

srfraser:
I like the idea, and I'm on board with the benefits.

From the description under option 1, it seems like we'd still be treating the individual scriptworker types as separate deployments, is that correct? For example, copying the release-services model, the current `scriptworker` would end up under `lib` and the individual workers would be under `src`. I don't have a strong opinion here, I'm just trying to make sure I'm clear about the proposal.

Would the deployment options be under the individual scriptworker directories, too, or separated out so there's one place to look to find out what options/permissions/etc one of them has? If we do, this would make it easier to have development versions. Having -prod, -dev and -other deployment configs might also help with the applescript situation.

Option 2 does mean we don't have to discuss the repository name, so that's an advantage ;) If we do need a name for it, though, `mozilla-releng/scriptworkers` is probably simplest, although I'm in favour of removing 'script'

I _imagine_ we're not at increased security risk from having the deployment & tooling all in one place with the same set of people having access, but it would be good to have more informed opinions about that. Does access to update and deploy one component imply access to them all? We probably have that scenario now, but through inertia rather than conscious choice.

Would we want to adjust the names for the workerTypes, and have 'v2' everywhere?

mitchhentges:
> Would we want to adjust the names for the workerTypes, and have 'v2' everywhere?

Since the underlying worker behaviour is still the same, I'm don't think that there's a breaking change in behaviour - therefore, I don't think bumping the version is necessary.

escapewindow:
> From the description under option 1, it seems like we'd still be treating the individual scriptworker types as separate deployments, is that correct?

I believe that in both options, we should be able to deploy each \*script individually, though it may become common to deploy them all concurrently.

> For example, copying the release-services model, the current `scriptworker` would end up under `lib` and the individual workers would be under `src`. I don't have a strong opinion here, I'm just trying to make sure I'm clear about the proposal.

Under this proposal (both options), scriptworker itself will remain in its own repo, and the \*scripts repo will pull it in as a python dependency. We will have `scriptworker.client` code moved into a `scriptworker-client` directory and module, and each of the \*scripts can use it as a dependency. (In the second option, the `scriptworker.client` code may move to a `lib/scriptworker_common` directory to follow naming conventions.)

> Would the deployment options be under the individual scriptworker directories, too, or separated out so there's one place to look to find out what options/permissions/etc one of them has? If we do, this would make it easier to have development versions. Having -prod, -dev and -other deployment configs might also help with the applescript situation.

Most likely the individual \*script directories would have all we need to test or deploy them as python modules, as well as any docker-specific config. The overall automation will probably be in .taskcluster.yml and in some shared area. Since we're moving to nix as our solution, we most likely will follow any conventions as needed for nix.

> Option 2 does mean we don't have to discuss the repository name, so that's an advantage ;) If we do need a name for it, though, `mozilla-releng/scriptworkers` is probably simplest, although I'm in favour of removing 'script'

Since `scriptworker` isn't moving, it may be confusing to have 2 separate repos with an `s` being the only differentiator.

> I _imagine_ we're not at increased security risk from having the deployment & tooling all in one place with the same set of people having access, but it would be good to have more informed opinions about that. Does access to update and deploy one component imply access to them all? We probably have that scenario now, but through inertia rather than conscious choice.

We have all the deployment & tooling in one place in puppet at the moment, so this isn't a backwards step. Agreed this is a concern, and one that we've discussed with the security team. We will need to balance security concerns with operational concerns.

We will probably have rules about what can deploy when, and put them into the monorepo, governed by review. We will add a dockerhub scriptworker that verifies that any publishing request comes from the official branch(es) off the official repo before publishing.

> > Would we want to adjust the names for the workerTypes, and have 'v2' everywhere?
> 
> Since the underlying worker behaviour is still the same, I'm don't think that there's a breaking change in behaviour - therefore, I don't think bumping the version is necessary.

+1. If our assumptions about a breaking change in behavior are wrong, `v2` is a fine option.

mitchhentges:
I'm personally leaning towards option 1:
* I think that since all the `*scripts` are so similar (fulfill the same use case, need to integrate with the same thing - `scriptworker`), there's a lot of good reason to have them together. My favourite outcome of sharing a repository is that, when shared code changes happen (e.g.: `scriptworker` client library API change), you can make those changes all together efficiently.

* From what I know so far, there's far less in common between the *scripts and the things in `release-services`. The two similarities that were noted were that they both (do/will) need `papertrail` and `nix`. IMO, sharing the same dependencies/build target isn't as strong of a reason to share code in a monorepo. Will there frequently be the kind of changes that would need to propagate through both `release-services` projects and `*scripts`?

Maybe a bad analogy, but perhaps this is kind of like if there's multiple projects that target Android: they'll both have to define their Android build, and they might share similarities (target the same platform, use the same version of `gradle`). However, for Android projects, there's enough of a need for de-coupling that they usually don't get mono-repo'd

* I'm also a bit nervous about forcing what _could_ be an unrelated/deferred change onto `*scripts` earlier than needed: If `please` is still a little rough around the edges, perhaps it would be get to let it stabilize more before having these edges affect the `*scripts`?

-----

Perhaps a good corollary would be [`android-components`](https://github.com/mozilla-mobile/android-components/): all these separate components that fulfill the same interface sit in a single monorepo, while projects that share less significant similarities remain in separate repos (like [`focus`](https://github.com/mozilla-mobile/focus-android/) and [`fenix`](https://github.com/mozilla-mobile/fenix))

-----

Finally, a disclaimer: I'm not familiar with `nix`, so there might be a special limiting factor that really encourages sharing all projects that are `nix`-ed into a single repo. :)

garbas:
> I'm personally leaning towards option 1:
> 
>     * I think that since all the `*scripts` are so similar (fulfill the same use case, need to integrate with the same thing - `scriptworker`), there's a lot of good reason to have them together. My favourite outcome of sharing a repository is that, when shared code changes happen (e.g.: `scriptworker` client library API change), you can make those changes all together efficiently.
> 
>     * From what I know so far, there's far less in common between the *scripts and the things in `release-services`. The two similarities that were noted were that they both (do/will) need `papertrail` and `nix`. IMO, sharing the same dependencies/build target isn't as strong of a reason to share code in a monorepo. Will there frequently be the kind of changes that would need to propagate through both `release-services` projects and `*scripts`?
> 
> 
> Maybe a bad analogy, but perhaps this is kind of like if there's multiple projects that target Android: they'll both have to define their Android build, and they might share similarities (target the same platform, use the same version of `gradle`). However, for Android projects, there's enough of a need for de-coupling that they usually don't get mono-repo'd
> 

I think the point in having a monorepo is not because things can not be shared otherwise. In monorepo projects are still going to be individually redistributable as before.

We all want to follow best practices, but enforcing this is very hard if not impossible. Having everything in one place makes this things at least possible if not even easy to do. Looking at release-services repo and how it all started I knew that we are going to share code, but I wouldn't expect that much code being shared between projects as we have. Keeping all projects up to date with our releng python guidelines is easy, since when one project solves the problem all projects immediately (in the same commit) benefited.

How much our projects share is very much a cultural thing and not so much technical.

>     * I'm also a bit nervous about forcing what _could_ be an unrelated/deferred change onto `*scripts` earlier than needed: If `please` is still a little rough around the edges, perhaps it would be get to let it stabilize more before having these edges affect the `*scripts`?
> 

As I already said at the meeting and many other times. Nobody is forcing `./please` command on anybody. All python projects are installable as before with the tools everybody is comfortable with. `./please` command is there to provide a convenience for those who need it. And a tool is not going to be improved if we are not going to use it, that is for sure.

> Perhaps a good corollary would be [`android-components`](https://github.com/mozilla-mobile/android-components/): all these separate components that fulfill the same interface sit in a single monorepo, while projects that share less significant similarities remain in separate repos (like [`focus`](https://github.com/mozilla-mobile/focus-android/) and [`fenix`](https://github.com/mozilla-mobile/fenix))
> 
> Finally, a disclaimer: I'm not familiar with `nix`, so there might be a special limiting factor that really encourages sharing all projects that are `nix`-ed into a single repo. :)

Nothing forces this in Nix, you can have any it anyway you want. The only limitation is that you need to know all of your dependencies (and their URL with sha256) before you run nix commands. Now some call this limitation, I call this the only sensible way to work :)

@escapewindow I think a third options is missing. We can use nix (and `./please` command) outside release-services repo as well. So like and option between 1 and 2.

escapewindow:
> @escapewindow I think a third options is missing. We can use nix (and `./please` command) outside release-services repo as well. So like and option between 1 and 2.

I think this is the direction we're going.

- scriptworker-scripts as its own repo
  - (these still might migrate to release-services in the future, if it makes sense; not ruling that out)
- we support publishing to pypi, but also support moving each of these \*scripts to nix+docker, starting with shipitscript. We can copy the `please` command if that's what the people porting to nix want to use.

tomprince:
> we support publishing to pypi

Given that we don't currently publish *scripts to pypi, I suspect this isn't necessary.

escapewindow:
> > we support publishing to pypi
> 
> Given that we don't currently publish *scripts to pypi, I suspect this isn't necessary.

We might need it for the scriptworker-client module in this repo. Possibly not.

rfcs/0008-scriptworker-script-monorepo.md
escapewindow: We seem to be leaning towards a separate repo, with the capability of merging later if we want.

rfcs/0008-scriptworker-script-monorepo.md
escapewindow: I'm going with the fewest overall changes for this rfc (semver, \*script-only monorepo). Let's move forward with that, and we can move towards changes like towncrier in followups.

We may or may not use nix. If so, we have shipitscript as a model. If not, we'll still be moving to docker and gcp. Let's figure this out as its own question; we don't have to figure that out here.

escapewindow:
We're about to start using the POC [scriptworker-scripts repo](https://github.com/escapewindow/scriptworker-scripts) for iscript. We'll adjust this as needed/wanted, and then we can add more *scripts to it.

rfcs/0008-scriptworker-script-monorepo.md
```diff
@@ -0,0 +1,53 @@
...
+Initially, we can cover automated testing via a `.travis.yml` that covers the entire repo, but this will be complex enough that we will want a `.taskcluster.yml` to test each \*script concurrently, and potentially build wheels and docker images in automation.
+
+Any change in shared behavior should span all \*scripts as applicable.
+
+We can do things like pyup dependency pinning across the repo like we do with puppet.
+
+We need to have ways of excluding or special-casing \*scripts. `balrogscript` is py2; the upcoming `iscript` will be targeted towards running on mac hosts instead of deployed through kubernetes.
+
+We should version each module or script like we do currently: semver. Libraries (e.g., scriptworker-client) will bump major versions every time they make a non-backwards-compatible change. The \*scripts will bump their major versions when they make a task- or puppet- (runtime- or deployment-) backwards-incompatible change.
```
rail: We should also unify where we store the version information (`version.txt` vs `_version.py`)

rfcs/0008-scriptworker-script-monorepo.md
```diff
@@ -0,0 +1,53 @@
...
+
+(Coop tweeted a [thread about taskcluster going to a monorepo](https://mobile.twitter.com/ccooper/status/1081264280429871104). We have a similar situation: small code size, small team size, and we're also going to be handing off operational responsibilities to cloudops.)
+
+# Details
+
+We create a github repo, `mozilla-releng/scriptworker-scripts`. We move the code from `scriptworker.client` into a `scriptworker-client` subdirectory, with its own `setup.py`. Tests will go under each subdirectly, e.g. `scriptworker-client/tests/` or `signingscript/tests`. We can move each production-ready, releng-maintained \*script into the monorepo over time.
+
+Each \*script will have its own `setup.py`.
+
+Initially, we can cover automated testing via a `.travis.yml` that covers the entire repo, but this will be complex enough that we will want a `.taskcluster.yml` to test each \*script concurrently, and potentially build wheels and docker images in automation.
```
rail: We should also consider using `taskgraph` and decision task based approach, so we don't run all the builds/tests/deployments every single time. Maybe use it from the beginning?

rfcs/0008-scriptworker-script-monorepo.md
```diff
@@ -0,0 +1,53 @@
...
+
+Any change in shared behavior should span all \*scripts as applicable.
+
+We can do things like pyup dependency pinning across the repo like we do with puppet.
+
+We need to have ways of excluding or special-casing \*scripts. `balrogscript` is py2; the upcoming `iscript` will be targeted towards running on mac hosts instead of deployed through kubernetes.
+
+We should version each module or script like we do currently: semver. Libraries (e.g., scriptworker-client) will bump major versions every time they make a non-backwards-compatible change. The \*scripts will bump their major versions when they make a task- or puppet- (runtime- or deployment-) backwards-incompatible change.
+
+# Open Questions
```
rail: I'm not sure how to properly release and deploy (assuming we will deploy from CI) using github releases. When you create a github release it requires a tag, which doesn't make too much sense with multiple versions in the repo.

Or maybe we don't use the github release feature and use a tool to start release CI tasks.

Another option would be using the "shippable builds/artifacts" approach. We generate ready-to-be-deployed artifacts and have a process to promote them.

escapewindow:
@mozilla-releng/releng , any further comments before we merge this? The existing monorepo is in good working order, and any future changes and improvements can proceed from there.

JohanLorenzo:
:shipit:

PR #13, opened 2019-01-28
Proposal to not use balrog for fennec

Callek:
I am proposing that we do not use Balrog for Fennec going forward

[Rendered](https://github.com/Callek/releng-rfcs/blob/no_fennec_balrog/rfcs/0013-disable-fennec-balrog.md)

### DIFF
--- a/rfcs/0013-disable-fennec-balrog.md
+++ b/rfcs/0013-disable-fennec-balrog.md
@@ -66,13 +66,13 @@ These numbers also greatly reduce in quantity when including only recent version
 Unless no new information as to reasons we shouldn't do this, I will be going ahead with implementation with intended landing
 on Feb 5, 2019.
 
+* Investigate if Fennec nightly updates are still needed in Balrog
+   * [Bug 1387643](https://bugzil.la/1387643)
 * Disable Balrog Submission of new builds for all channels
-   * Bug XXXX
+   * [Bug 1524394](https://buzgil.la/1524394)
 * Disable in-product updater for all version of Fennec
-   * Bug XXXX
+   * [Bug 1520248](https://bugzil.la/1520248)
 
-* <link to tracker bug, issue, etc.>
-* <...>
 
 # FAQ
 
--- a/rfcs/0013-disable-fennec-balrog.md
+++ b/rfcs/0013-disable-fennec-balrog.md
@@ -1,0 +1,86 @@
+# RFC 13 - Proposal to Disable Fennec Balrog Submission
+* Comments: [#13](https://github.com/mozilla-releng/releng-rfcs/issues/13)
+* Proposed by: @Callek
+
+# Summary
+
+I am proposing that we disable Fennec Balrog submission, which causes there to be no offered updates for new
+Fennec builds when Fennec has been sideload installed.
+
+With this work we will do all of the following:
+
+* Disable Updates on **Single Locale** Fennec builds on `nightly` channel
+* Disable Updates on **All** Fennec builds on `nightly` channel
+* Disable Updates on **All** Fennec builds on `beta` channel
+* Disable Updates on **All** Fennec builds on `release` channel
+
+## Motivation
+
+There are a handful of configuration differences and bugs that make Sideload Install updates difficult to manage, both
+from a product perspective and from the Release Engineering side. Explicitly no longer supporting automatic updates
+via Sideload install will simplify this work.
+
+I came across a few bugs resulting in single locale and non-en-US multi locale updates being busted on Fennec
+Nightly via Balrog right now. This has seemingly gone unnoticed so user impact appears to be low. More
+details on those bugs and current user metrics are highlighted below:
+
+* (***non**-en-US/multi*) updates via the channel `nightly-mozilla-central` which Balrog does not currently recognize,
+  would need to change to `nightly` for product and rescue the users in balrog
+     * <s>Fix ETA < 2 days of work.</s>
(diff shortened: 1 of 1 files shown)


Callek:
> Will we want to somehow inform users of sideloaded Firefox that they will no longer receive updates? I'm not sure that they'll all see the FAQ section of this RFC :)

I'm not sure there is any way to give mobile users a `detailsURL` or a `showURL` directive, doesn't look like anything in Fennec's update code handles that.

The code that queries balrog is at https://searchfox.org/mozilla-central/source/mobile/android/base/java/org/mozilla/gecko/updater/UpdateServiceHelper.java#166 and it looks like it just returns a url for an available update, and that is it.

theres-waldo:
Is there a significant cost to keeping this in place if we do not fix the mentioned bugs?

As someone who tries to avoid Google accounts but runs Fennec Nightly for testing and dogfooding purposes, I've been relying on this auto-update mechanism. Manually updating each time would be significantly less convenient.

I also feel that from a policy point of view, we should be moving in the other direction and improving the experience for users who do not use Google accounts (e.g. fixing [bug 1192279](https://bugzilla.mozilla.org/show_bug.cgi?id=1192279)).

Callek:
> Is there a significant cost to keeping this in place if we do not fix the mentioned bugs?

For one thing if we want locales other than English to also update we need to fix the single locale bug. We have the issue that users who install en-US get a *different* build type entirely. We don't have any formal testing of the update paths involved, and we don't have good code ownership or stewardship of these update codepaths.

It also adds a lot of cognitive overhead when dealing with Fennec code as well as when dealing with release automation. Having release automation submit to Balrog when product doesn't use them for beta/release is wasteful in many areas.

Lastly users who allow this to update on their device are prone to possible other issues because of the need to disable several Android Security settings. Even technical users could get into trouble with those settings disabled.

> As someone who tries to avoid Google accounts but runs Fennec Nightly for testing and dogfooding purposes, I've been relying on this auto-update mechanism. Manually updating each time would be significantly less convenient.

I agree, but the overall usage numbers make me less eager to keep this for the added complexity and frustration. Google Play has a policy that apps shipped on it cannot self-update, and I am unaware of any other apps than Fennec that do self-update like this. I suspect that, on Android, using the App Stores will be the way that most apps should update, whether that be the Google Store, or something else.

The fact that we don't support updates via this mechanic on the Release channel makes me even less concerned, since those users can't even use our builds in a sideload fashion without also incurring the update pain themselves.

> I also feel that from a policy point of view, we should be moving in the other direction and improving the experience for users who do not use Google accounts (e.g. fixing [bug 1192279](https://bugzilla.mozilla.org/show_bug.cgi?id=1192279)).

Should product decide to revisit this and decide that sideload updates are a thing they want to support for fennec in a more formal way, I am willing to revisit this, I don't suspect adding support anew at a later date would be too onerous, and would additionally allow us to rethink how some of this stuff works on Android.

buttercookie42:
> Lastly users who allow this to update on their device are prone to possible other issues because of the need to disable several Android Security settings. Even technical users could get into trouble with those settings disabled.

It's only *one* setting, and with Android O the whole thing [was completely revamped anyway](https://android-developers.googleblog.com/2017/08/making-it-safer-to-get-apps-on-android-o.html) - the whole thing now works like all other Android runtime permissions, i.e. instead of globally allowing sideloading, you now authorise individual apps as to whether they should be allowed to install other packages or not.

Callek:
No blocking issues that were not forseen, calling this happening.

LegNeato:
This seems extremely shortsighted.

A large percentage of Android users do not have Google Play and do not have stores that Mozilla ships to (though we have discussed using my company https://pocketship.com to automatically ship to some of them). Though some third party stores grab Firefox and put it in their store, they basically only do it once and it is an old version.

All growth in mobile devices is happening in India, China, and Africa. Sideloading via wifi or bt sharing is large in all those places and we'd want to keep Firefox updated when possible.

Additionally, as preload deals get more and more important to Mozilla it will be more critical to rely on our own updater...preloaders usually qualify with one Firefox version and never touch it again. Seeing how the "preloaded on device" set may not intersect with the "device that has store with latest Firefox", it seems prudent to keep the updater on.

If you have the updater you are not beholden to other companies to ship your product. Relying on the neutrality of Google, Amazon, Tencent, Yandex, and others seems naive. Even if they never pull the rug out from under Firefox, having control of your own updates means you can do interesting things like ship test versions more quickly to a small subset of the population using targeting more specific than the stores give you.

I don't find the motivation section compelling. There is a strategic reason why Facebook did their own updates even though they were in Google Play, and a reason why Facebook has an installation daemon they control on every Samsung device...and those reasons apply to Mozilla as well.

Callek:
@LegNeato, thank you for this feedback. I wonder if there may be some avenues of the decision process on my end that are not clear, or I may be misunderstanding some of your points. I'll reach out to you directly in e-mail over the next business day or two and then circle back to this PR to dump the outcome of our conversation.

Callek:
Circling back as promised,

So @LegNeato is happy with my e-mailed restating and appears happy.

The key points are:
* we are not removing Android support from Balrog itself.
* This was prompted primarily by a chunk of infrastructure (CI) costs that are only useful for an extremely small portion of Nightly-Only users (release and beta don't get any benefit currently)
* To support Balrog updates for any android product, we should get some actual product level prioritization around supporting some of the must-have features of Balrog (such as a desupport page, or the ability to show a what's new page, etc.) since the implementation is completely removed from Desktop's implementation.

Should Product come back and say they want to use Balrog (or any update service releng could provide) we will discuss and make that happen, but such a desire would come with it a commitment by product teams to support it at the product level.

PR #14, opened 2019-02-01
Creates RFC for bumping script versions to at least 1.0.0

mitchhentges:
[rendered](https://github.com/mitchhentges/releng-rfcs/blob/script-versions-1.0.0-production/rfcs/0014-use-production-versions-for-scripts.md).
Based on the discussion from [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1524248)

### DIFF
--- a/rfcs/0014-use-production-versions-for-scripts.md
+++ b/rfcs/0014-use-production-versions-for-scripts.md
@@ -1,22 +1,22 @@
-# RFC 14 - Use production versions for scripts
+# RFC 14 - Use production semver versions for releng packages
 * Comments: [#14](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/14)
 * Proposed by: @mitchhentges, on behalf of discussion [in bug 1524248](https://bugzilla.mozilla.org/show_bug.cgi?id=1524248) 
 
 # Summary
 
-Bump all script versions to >= `1.0.0`.
+Bump all releng library and *script versions to >= `1.0.0`.
 
 ## Motivation
 
-We're using our *scripts in production. According to [semver.org](https://semver.org/#how-do-i-know-when-to-release-100),
+We're using our packages in production. According to [semver.org](https://semver.org/#how-do-i-know-when-to-release-100),
 if software is being used in production, it should be >= `1.0.0`.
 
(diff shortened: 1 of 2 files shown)


mitchhentges:
Pinging @escapewindow and @JohanLorenzo, since y'all were fielding the discussion in that ticket

escapewindow:
The scripts' versions are less important, since they're not used as libraries, although I believe Tom suggested we bump the major version if we are non-backwards compatible with the current puppet configs. Modules we release as libraries are more important.

rfcs/0014-use-production-versions-for-scripts.md
```diff
@@ -0,0 +1,23 @@
+# RFC 14 - Use production semver versions for releng packages
+* Comments: [#14](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/14)
+* Proposed by: @mitchhentges, on behalf of discussion [in bug 1524248](https://bugzilla.mozilla.org/show_bug.cgi?id=1524248) 
+
+# Summary
+
+Bump all releng library and *script versions to >= `1.0.0`.
```
escapewindow: I'm wondering if this is overly broad? We went from a discussion about mozapkpublisher, which I believe tries to adhere to semver, to applying semver rules to all releng libraries and \*scripts regardless of usage.

We had a [discussion](https://github.com/mozilla-releng/releng-rfcs/pull/8#discussion_r248868038) in the monorepo rfc about versioning; we may ditch it when we move to nix deployment, stay with semver, or go with calver. I'm also not sure what's happening in services land, or other one-off libraries that aren't in use in scriptworker.

At the very least, we should maybe add `for [production] releng packages` or `Bump all [production] releng ...` to clarify that we don't have to bump in-progress, not-yet-live packages.

mitchhentges: Ah, great point! I was assuming that all of our packages right now are production :)

lundjordan:
@mitchhentges - okay to close this and add it to the repo? Sounds like we have consensus or at least no outright objection?

mitchhentges:
I suppose that after a month of being in "final comment phase", we _might_ be ok to merge this :)

PR #21, opened 2019-04-03
Add RFC for moving Balrog's Releases History to Google Cloud Storage.

mozbhearsum:
Rendered version: https://github.com/mozbhearsum/releng-rfcs/blob/balrog-releases-history/rfcs/0021-balrog-releases-history-gcs.md

### DIFF
--- a/rfcs/0021-balrog-releases-history-gcs.md
+++ b/rfcs/0021-balrog-releases-history-gcs.md
@@ -1,5 +1,5 @@
-# RFC <number> - Move Balrog's releases_history table to Google Cloud Storage
-* Comments: [#<number>](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/<number>)
+# RFC 21 - Move Balrog's releases_history table to Google Cloud Storage
+* Comments: [#21](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/21)
 * Proposed by: @mozbhearsum
 
 # Summary
--- a/rfcs/0021-balrog-releases-history-gcs.md
+++ b/rfcs/0021-balrog-releases-history-gcs.md
@@ -47,28 +47,28 @@ The simplest way to understand the proposed changes is to look at the flow of da
                           +---------+
 ```
 
-With the proposed changes, only writes go through the admin backend, and are ultimately handled by the Agent. Anything wishing to do reads, does so directly with GCS:
+With the proposed changes, only writes go through the admin backend. Anything wishing to do reads, does so directly with GCS:
 ```
 +---------------+
 |               |
-|    Balrog     +----+                         Reads and Writes
-| Scriptworkers |    | Updates                 Release History to
-|               |    | Releases    +---------+ a Temporary Table +---------+
-+---------------+    +------------>+         +------------------>+         |
-                     +------------>+  Admin  |                   |  MySQL  |
-+---------------+    |             |         +<------------------+         |
-|               |    |             +----+----+                   +---------+
+|    Balrog     +----+                        
+| Scriptworkers |    | Updates               
+|               |    | Releases    +---------+
++---------------+    +------------>+         |
+                     +------------>+  Admin  |
++---------------+    |             |         |
+|               |    |             +----+----+
 |   Frontend    +----+                  |
 |               |                       |
-+-------+-------+                       | Reads Release History
-        ^                               | from Temporary Table
++-------+-------+                       |                      
+        ^                               |                     
         | Reads Release History         |
-        |                               v
-   +----+----+                     +----+----+
-   |         |                     |         |
-   |   GCS   +<--------------------+  Agent  |
-   |         | Writes Release      |         |
-   +---------+ History             +---------+
+        |                               |
+   +----+----+                          |     
+   |         |                          |     
+   |   GCS   +<--------------------------     
+   |         | Writes Release                 
+   +---------+ History                            
 
 ```
 
@@ -89,8 +89,7 @@ Local development environments will *not* have their own buckets by default (mor
 
 ## Balrog Changes Needed
 
-* Replace writes to `releases_history` table with writes to a temporary table.
-* Enchance the Agent to monitor for these rows and write them to GCS.
+* Replace writes to `releases_history` table with writes to GCS.
 * Numerous changes to the Frontend:
     * Query GCS when Release history is requested instead of admin. The appropriate bucket can be queried by prefix to find all versions of the Release.
     * Change the way we do reverts. Instead of calling a special admin endpoint that copies data from `releases_history` to `releases`, we'll simply have the Frontend update a Release with the old data (or schedule a change).
@@ -109,8 +108,6 @@ Local development environments will *not* have their own buckets by default (mor
 
 # Open Questions
 
-* Will access to Release History actually be faster? Need to compare requests to the admin app vs. requests to GCS to verify.
-
 # Implementation
 
-TODO
+https://github.com/mozilla/balrog/pull/910
--- a/rfcs/xxxx-balrog-releases-history-gcs.md
+++ b/rfcs/xxxx-balrog-releases-history-gcs.md
@@ -1,0 +1,116 @@
+# RFC <number> - Move Balrog's releases_history table to Google Cloud Storage
+* Comments: [#<number>](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/<number>)
+* Proposed by: @mozbhearsum
+
+# Summary
+
+Move Balrog's `releases_history` table from its current location as a table in the MySQL database, and into Google Cloud Storage (GCS).
+
+## Motivation
+
+`releases_history` is an extremely large, infrequently used table (~450GB, more than 99% of Balrog's database). Migrating it to GCS has a number of benefits:
+* Ease our migration of Balrog into Google Cloud Provider (GCP), because the database will become small enough that downtime or disabling writes will be a non-issue.
+* Simplify retention with GCS Lifecycle Policies. These would fully replace our Python code that currently purges data from `releases_history`.
+* Speed up access to Release History, because GCS access is faster than going through the admin backend and database. (see open question about this)
+* Allow for local Balrog installations to access full Release history.
+
+# Details
+
+## Architecture
+
+The simplest way to understand the proposed changes is to look at the flow of data now, and how it would look with the proposed changes. In the current world, everything that reads or writes Releases history goes through the admin backend to do so, as shown here:
+```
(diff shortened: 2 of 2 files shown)


mozbhearsum:
Rendered version is at https://github.com/mozilla-releng/releng-rfcs/blob/7f0651a527537cd4d158095731e5116d1427d660/rfcs/0021-balrog-releases-history-gcs.md

Flagging a few people that are most likely to have comments: @catlee, @nthomas-mozilla, @oremj, @jbuck

rfcs/0021-balrog-releases-history-gcs.md
```diff
@@ -0,0 +1,116 @@
...
+    * An example of a full key name (with folder) is: Firefox-66.0.2-build1/547-1553640087550-balrog-ffxbld.json
+* Key values will be the Release `data`.
+
+The dev and stage enviroments will have their own sets of buckets. It's important that these environments are able to fully simulate production, so sharing the production buckets is inappropriate here. However, we don't need to care about long term retention of any of their data, so they will be configured to retain it for only 5 days (including for non-nightlies).
+
+Local development environments will *not* have their own buckets by default (more on that in the Caveats section).
+
+## Balrog Changes Needed
+
+* Replace writes to `releases_history` table with writes to a temporary table.
```
oremj: Might be better to write this to pubsub or just directly to GCS avoiding temp tables completely.

rfcs/0021-balrog-releases-history-gcs.md
```diff
@@ -0,0 +1,116 @@
...
+
+## GCS
+
+Buckets will be configured as follows:
+* One bucket for Nightly Releases, with a 14 day Lifecyle Policy for all keys
+* One bucket for all other Releases, with no Lifecycle Policy (ie: they will be kept indefinitely)
+* Both buckets will be made publicly available, as Releases contain no secrets
+* We will create a folder at the top level of the appropriate bucket for each Release. Eg: Firefox-66.0.2-build1
+* Within each folder, there will be a key per data version with basic data (data version, timestamp, changed_by) encoded in the name
+    * An example of a full key name (with folder) is: Firefox-66.0.2-build1/547-1553640087550-balrog-ffxbld.json
```
oremj: How will links to these files be created? GCS doesn't offer a browser based index listing, except for within the admin console.

mozbhearsum: We won't create any links. Anybody that needs them will know which release name they want history of, and can query by prefix for that. Eg: https://www.googleapis.com/storage/v1/b/bhearsum_balrogtest_releases_history/o?prefix=Firefox-66.0.2-build1/

rfcs/0021-balrog-releases-history-gcs.md
mozbhearsum: Hmm, the main reason I was looking at temp tables and having the Agent do it was to avoid adding latency to writes (which have a significant impact to the number of update races/retries that clients hit), and writing to a temp table would roughly equivalent to the current releases_history (if not faster). Perhaps it's worthwhile doing a few tests to see if writing to pubsub or GCS directly would be faster...

mozbhearsum: I did some testing, and It actually looks like MySQL might be the fastest for writes? I used a GCE instance to try writing to a CloudSQL database (initialized with a Balrog schema), uploading to GCS, and publishing a message to pubsub. I ran each test 10 times with a ~2.8MB piece of data (this is not an abnormal row size for `releases_history`) and got these results:
* GCS Writes (done with `gsutil`): ~800ms
* MySQL insert (done on the command line, reading an INSERT statement from a file): ~200ms
* Pubsub publishing (done with cURL + base64 encoded data (making it ~3.6MB): ~400-500ms

There may be holes in my methadology, so I'd be happy for someone to cross check.

oremj: I tested GCS directly with the following python script.

```python
import time

from google.cloud import storage

storage_client = storage.Client()

bucket_name = "balrog-speed-test"

bucket = storage_client.get_bucket(bucket_name)

content = b'a' * 2800000

timings = []

for i in range(2000):
    start = time.time()
    blob = bucket.blob("test2/test" + str(i))
    blob.upload_from_string(content)
    timings.append(time.time() - start)


avg = sum(timings) / len(timings)
median = timings[int(len(timings) / 2)]
print("Avg:", avg)
print("Max:", max(timings))
print("Median:", median)
print("Min:", min(timings))
```
and got the following results:

oremj@balrog-speed-test:~$ python3 gcstest.py
Avg: 0.3643922011852264
Max: 0.9011557102203369
Median: 0.38791489601135254
Min: 0.2687060832977295

oremj: Also tested pubsub:
```python
import time

from google.cloud import pubsub_v1

publisher = pubsub_v1.PublisherClient()

topic_path = publisher.topic_path("oremj-sandbox", "balrog-speed-test")

content = b'a' * 2800000

timings = []

for i in range(2000):
    start = time.time()
    future = publisher.publish(topic_path, data=content)
    future.result()
    timings.append(time.time() - start)


avg = sum(timings) / len(timings)
median = timings[int(len(timings) / 2)]
print("Avg:", avg)
print("Max:", max(timings))
print("Median:", median)
print("Min:", min(timings))
```
Results:
```
Avg: 0.1633094903230667
Max: 0.4166691303253174
Median: 0.16144347190856934
Min: 0.1445479393005371
```

~160ms pubsub publishing seems pretty good.

oremj: And MySQL too:
```python
import time

import MySQLdb

db = MySQLdb.connect()
c = db.cursor()

content = b'a' * 2800000

timings = []

for i in range(2000):
    start = time.time()
    res = c.execute("""INSERT INTO blob_test VALUES (%s)""", (content, ))
    db.commit()
    timings.append(time.time() - start)


avg = sum(timings) / len(timings)
median = timings[int(len(timings) / 2)]
print("Avg:", avg)
print("Max:", max(timings))
print("Median:", median)
print("Min:", min(timings))
```
Results:
```
Avg: 0.5087654188871383
Max: 4.238947153091431
Median: 0.46433091163635254
Min: 0.0970454216003418
```

mozbhearsum: Huh, that's very interesting. I'm very surprised that your MySQL test is so much slower than mine. Perhaps that just means that the mysql command line client is much faster than MySQLdb. In any case, your tests are most likely more representative of what our performance would be like in Balrog itself, so it seems quite clear that pubsub has significantly faster writes than both GCS and CloudSQL.

oremj: @mozbhearsum Could be the instance size/storage I picked as well. I'll do one more test with a larger instance/more iops.

oremj: I re-ran the test with a much larger DB and got better results.
```
$ python3 mysqltest.py
Avg: 0.11573576951026916
Max: 0.6509559154510498
Median: 0.12839674949645996
Min: 0.08424234390258789
```

mozbhearsum: Hm, so that's pretty fast. Do we know if that would be representative of what we would see in production? Pubsub and CloudSQL being more than twice as fast as writing directly to GCS pushes me towards one of those options, and letting the Agent handle moving the data to GCS out of band.

Does this change your opinion/preference on where we write the data initially?

oremj: I'd prefer pubsub or direct to GCS, so we don't put additional strain on the database. Pubsub if we need to push response time and GCS directly if not.

Pubsub does have the limitation of max 10MB message size and would be an additional step.

Alternatively, can we spin off a thread or write to GCS asynchronously, so the responses don't block?

mozbhearsum: Something about using a thread feels pretty dirty to me - I guess it's very odd to have code associated with a request running after that request has returned.

I honestly don't know if the GCS write times are going to be an issue or not. It looks like ~1s response times for the requests I'm concerned about are pretty much the norm already. Writing directly to GCS looks like it would add about 30% to that (~400ms to write to it, minus ~100ms because we won't be writing it to MySQL anymore). That doesn't seem _too_ terrible, but it would probably cause additional requests to receive 400s due to losing a race, which has a bit of pile-on effect.

The pubsub limitations shouldn't be an issue - our messages would be nowhere close to 10MB now, or in the forseeable future. This option is looking more attractive, despite being a bit more work...

oremj: 400s due to all uwsgi workers being occupied? We could always add more capacity.

mozbhearsum: Ah, no. 400s due to data races. We often have tons of requests trying to update the same Release, and they have to retry if they race against one another in some circumstances. Mainly caused by our data model/architecture.

mozbhearsum: @oremj - could we bring up a realistic admin+database in GCP? I'd like to do some more testing to see if writing directly to GCS is a viable option.

oremj: We can use the environment that was spun up in https://github.com/mozilla-services/cloudops-infra/issues/797

mozbhearsum: Mind if I use the new stage instead? (That's where our submissions for things like this will eventually want to end up anyways.)

oremj: Yep, the stage environment works for me.

mozbhearsum: I finally got this test done against the new stage environment. With the current code (continuing to write to the `releases_history` table, the average release submission took ~400ms on average. Writing the data to GCS directly instead caused the requests to take ~700ms on average, with ~4x the retries.

In and of itself, I think that request time is too bad, but the number of retries indicates that the slower requests are causing more data races. That's also not necessarily awful, as long as things still complete in a reasonably timely manner, but it's hard to fully predict what will happen when we have many releases (or many nightlies) running at once. I _think_ it's probably going to be fine, since parallel releases don't collide with one another (just with other submissions to the same release), but I can't be sure.

Some ways we might be able to improve our odds:
- Bump up the instance size for admin (the load is very peaky, so this might have an outsized effect at peak times)
- Do something network related to speed up write time to gcs (I don't even know if this is possible)

mozbhearsum: After a bunch of testing, it looks like writing directly to GCS will be viable. I'll update the RFC to reflect this.

PR #22, opened 2019-04-22
Best Practices for RelEng Applications

mozbhearsum:
[Rendered RFC](https://github.com/mozbhearsum/releng-rfcs/blob/project-standards/rfcs/0022-app-best-practices.md)


rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+COPY ${project_dir}/setup.py /app
+COPY lib/ /app/lib/
+```
+
+And from the root of the `services` repo, we can use `docker build -t shipit_api -f src/shipit/api/Dockerfile .` to build it.
+
+# Open Questions
+
+* Does Taskcluster's version of Docker/docker-in-docker support multistage builds?
+* Decision tasks or `.taskcluster.yml` v1? Both support the task dependencies needed.
```
mitchhentges: Personally, I find `.taskcluster.yml` files more difficult to maintain, so I'd prefer a fatter decision task? :thinking: :man_shrugging:

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+* Who will port RelMan's `mozilla-releng/services` projects?
+* Provide skeleton examples for Python and Javascript applications (explicitly not a cookiecutter repo/tool)?
+* Write linter to ensure projects conform to spec?
```
mitchhentges: The linter sounds valuable, but I am confident that we can safely defer it and manually verify conformity in the meantime. If there's enough issues with conformity, then we can re-visit the idea of building the tool later?

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+## Dependency Management
+
+Python projects will specify their dependencies in three files (as is standard for `pip-compile-multi`):
+* `requirements/base.in` will contain direct dependencies required for running the application
+* `requirements/test.in` will contain direct dependencies required for running tests
+* `requirements/local.in` will contain `tox`
+
+When any of these files change, `pip-compile-multi` will be used to generate the full list of pinned dependencies. Pyup.io will be used to keep them up-to-date.
+
+Javascript projects will specify their dependencies in `package.json`. As is the usual for Javascript projects, direct dependencies will be listed in `dependencies` and direct test and local development dependencies will be listed in `devDependencies`. When either of this lists change, `yarn lock` will be used to generate the full list of pinned dependencies. Greenkeeper.io will be used to keep them up-to-date.
```
mitchhentges: `yarn` was originally created because `npm` was awful with lockfiles and slow, but `npm` has made some significant progress since.
Keep in mind that  `yarn` requires additional setup for devs (Install NodeJS, then have to manually install `yarn` on top - which is challenging in Linux if you install node via package manager because then `npm i -g ...` requires `sudo`, etc).
Perhaps we should stick to `npm` unless the benefits to `yarn` are still significant enough

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Testing
+
+### Unit Tests
+
+All testing will be done within Docker container.
```
mitchhentges: I'm not confident that leaning on Docker makes sense for all of our projects:

* What about in-IDE testing with a debugger? Is that possible to do through Docker, and if so, how difficult is it to set up?
    * If our tests always run in a homogeneous environment, like Docker, it's possible that over-time the tests will develop quirks that make them fail on developer machines (e.g.: if running tests in-IDE). When we still have tests run "natively", we're more likely to find out if tests are somehow becoming environment-/platform-specific.
* This would block developers from becoming productive until they learn how to run Docker (and our specific docker image) locally, which is a significant cost

I'm not against running Docker images in CI, I think that it's important for PR-tests/release operations to happen in a clean environment, but I'm nervous about enforcing too much on developers' environments

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+Use standard and well known tools & processes for development, testing, building, and deploying RelEng applications. Specifically:
+* Use Docker & Docker Compose for running applications and tests locally
+* Use language appropriate tools for testing and dependency management
+    * For Python, this means `pip-compile-multi`, `tox`, and Pyup.io
+    * For Javascript, this means `yarn`, `neutrino`, `jest`, and Greenkeeper.io
+* Use Taskcluster to run tests and builds in CI
+* Maintain `production`, `stage`, and `dev` environments per application
+    * `production`, which may be automatically deployed when a new Release is created, or deployed to manually
+    * `staging`, which will automatically deploy on pushes to the `master` branch of a repo
+    * `dev`, which will automatically deploy on pushes to the `dev` branch of a repo
```
mitchhentges: I think that having common terminology and expectations for environments is :ok_hand:, great call!

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Deployment
+
+Every application will have a `production`, `stage`, and `dev` environment.
+
+### Production
+
+Production builds will happen in response to Github `release` events.
+
+Python projects will push both a `${project}_production` and `${project}_${hash}` tag to Dockerhub. If the project autodeploys to production, it will do in response to the `${project}_production` tag. If it manually deploys, a bug will need to be filed that references the `${project}_${hash}` tag.
```
mitchhentges: Why do we need to push `${project}_${hash}` to docker hub?
Shouldn't only tagged releases go to `production`? How do you define a "manual" deployment?

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+
+Javascript projects will publish a built version of the app as a Taskcluster artifact. If the project autodeploys to production, it will also publish to S3. If it manually deploys, a bug will need to be filed that references the built app.
+
+### Stage
+
+Stage builds will happen in response to push events on the `master` branch. Once a build is completed, it will automatically deploy to the `stage` environment. For Python projects, this means pushing a `${project}_stage` tag to Dockerhub. For Javascript projects, this means publishing to S3.
```
mitchhentges: Can you have multiple docker images with the same tag (`${project}_stage`)?

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+
+The stage environment should be used as a final testing ground before pushing to production. Code typically bakes here for ~24h before moving to production.
```
mitchhentges: I'm not sure that it's true in the general case that code will bake in `staging` for ~24h before going to production - for example, for `mozapkpublisher` or `pushapkscript`, we frequently merge a PR, then immediately release it and consume it.

rfcs/0022-standard-tooling.md
mitchhentges: Are we deploying python projects to `pip` any more?

rfcs/0022-standard-tooling.md
mozbhearsum: Yeah, I don't have strong feelings here either. I'm not even certain we need to be prescreptive beyond "Use Taskcluster", although it doesn't hurt to provide a good example with `.taskcluster.yml` or a decision task.

rfcs/0022-standard-tooling.md
mozbhearsum: :+1: Probably not a blocker, but history has shown that we're terrible at manually maintaining things like this -- particularly for things that aren't under active development.

rfcs/0022-standard-tooling.md
mozbhearsum: I'm open to either. I picked `yarn` here because it's what I've seen being used more lately.

rfcs/0022-standard-tooling.md
mozbhearsum: The reason we want hash tags as well is because for things that do manual deployments (like Balrog), it's much clearer for us to tell CloudOps "please deploy tag X", where X is a specific, new tag. Maybe this is a historical artifact, though. I'd want to see what they have to say before doing away with them, though.

rfcs/0022-standard-tooling.md
mozbhearsum: Not in the same repository at the same time, no.

rfcs/0022-standard-tooling.md
mozbhearsum: Do you mean `pypi`?

rfcs/0022-standard-tooling.md
mozbhearsum: It's worth noting that this isn't meant to make it impossible to run tests directly on your own (I realize this isn't spelled out in the RFC, in retrospect), merely to make the *standard* way of running tests through Docker. This means that things such as in-IDE testing should continue to work.

I know this imposes a bit of a burden, but it also levels the playing field. We've had both employees and volunteers who run Windows in the past, and Docker is the only reasonable way that they can do anything locally. I believe we've also had certain circumstances where folks who develop on OS X end up with different results than Linux, and I think even issues with a Linux host environment being different to cause issues.

rfcs/0022-standard-tooling.md
mitchhentges: If we _did_ have a linter, and a project fell behind active development, doesn't that mean that the linter would only realize that our project is no longer compliant once development re-continues - at which point we'd notice that it's not compliant ourselves? The linter is interesting, but I'm not sure I follow quite how we'd want to apply it or the scope of the benefit we'll get from it :)

rfcs/0022-standard-tooling.md
mitchhentges: Fair enough :smile:
One other criteria to keep in mind is: `yarn` and `npm` will each have their own bugs that crop up once in a while. Regardless of the tool we standardize on, if we have issues with the other, we'll take longer to realize/address the issue with the non-standardized-on tool. Since `npm` is the "default" package manager in the ecosystem, I think that it's better to maximize stability with it instead of `yarn`.

rfcs/0022-standard-tooling.md
mitchhentges: Hmm, it true that we have developers of multiple platforms.
FWIW, I suppose that we have three options:

* On both CI and local dev environments, testing usually happens in Docker images. Devs can use outside-of-docker tooling, but they need to handle any machine-local quirks
    * Benefits: 
        * On any docker-compatible machine, tests will run, guaranteed :fire:
    * Downsides: 
        * Highly recommended to have docker to test software, so devs will need to learn how to run docker.
        * Running the software locally (not unit tests, but running as a CLI) is tough - for example, getting `scriptworker` to run with another script will require some Docker finaggling
        * Windows developers can't use their tools unless they update the entire project to be cross-platform
    * Notes: Our code doesn't become simpler by avoiding platform issues, because once someone runs machine-local tools and produces a patch addressing inconsistencies, we're going to want to share that patch for when someone else also needs to run machine-local tools. At this point, the cross-platform complexity has been re-introduced into the project
* Force Windows devs to use WSL (or alternative), make projects compatible with both Mac and Linux
    * Downsides:
        * Windows developers need to learn WSL to work on our code
        * Windows developers can't use their tools unless they update the entire project to be cross-platform
* Make projects cross-platform, patches must work cross-platform to be accepted
    * Benefits:
        * No extra tooling knowledge required for most project developers
        * On any support machine, tests will run, guaranteed :fire:
    * Downsides:
        * Need tooling to test our code on all supported development platforms (more maintenance)
    * Notes: This cause us to have more up-front work of cross-compatibility, but as mentioned in the notes for the docker-image solution, work for cross-compatibility would happen _eventually_ (or we aren't using machine-local tools, which may be hurting our productivity in other ways)

rfcs/0022-standard-tooling.md
mitchhentges: Err, I'm confused - I interpreted "hash" as the "commit hash" associated with the release.
If we're referring to naming releases after their version number (e.g.: `scriptworker_1.0.1`), then I'm 100% chuffed (I hope I'm using the term "chuffed" right: @s.rfraser).

IMHO having the release number/tag in the name better conveys "this is a production release"

rfcs/0022-standard-tooling.md
mitchhentges: Oh, I miscommunicated - I was trying to ask if it was possible to upload two docker images with the same tag (e.g.: are tags immutable), but I suppose they aren't, considering that `python:latest` exists :)

rfcs/0022-standard-tooling.md
mitchhentges: :man_facepalming: err, yeah, `pypi`.

> Are we still packaging our python projects and deploying them to `pypi` anymore, or are we using docker images instead?

rfcs/0022-standard-tooling.md
mozbhearsum: This is all pretty hand-wavey at this point, but I imagine the linter would be a status check that runs on pull requests, so any time the code changes we'd verify it. So if we change what the linter does, we wouldn't catch new issues until next time the code changes in a case like that, I think.

The main idea is simply to have something that helps ensure we don't accidentally (or purposely) go against our own standards -- because I expect us humans to not be as good at catching those sort of things.

rfcs/0022-standard-tooling.md
mozbhearsum: It's not specifically part of this proposal (although maybe it ends up being a side effect), but I think I've heard we're planning to stop publishing/relying on the public `pypi` for our own packages.

mozbhearsum: Scriptworkers are an interesting case here, I suppose, since they need Python packages published to our internal pypi mirrors in order to deploy, which I think is what you're getting at? I wonder if that will be the case when we move them to GCP. @rail may know.

rfcs/0022-standard-tooling.md
mozbhearsum: On cross platform support - I actually think it's an anti-feature to support any platforms that we don't need to in our tools and applications (especially Windows). With a few exceptions, we don't build general purpose things - we build specific things to solve specific problems, and they are generally deployed in a specific way that we are in control of. Any time spent ensuring that other configurations work is unlikely to pay off in production, and I don't think we should bother.

mozbhearsum: > On any support machine, tests will run, guaranteed

I don't think this will actually be the case. You still have system level dependencies to worry about, which will differ on different OS' and distributions. (This is one of the big things that Docker solves).

mitchhentges: > I actually think it's an anti-feature to support any platforms that we don't need to in our tools and applications (especially Windows)

I agree that including support for platforms that we don't support adds unnecessary logic to our projects.
However, the following two ideas are contradictory:
1. Don't include code in our projects that is specifically for supporting an environment other than our docker image
2. Allow developers to use machine-local tools

Our machine environments will differ from the docker image. How do you see us keeping our machine-local tools working if we can't merge our platform-specific fixes into the projects?

----

> > On any support machine, tests will run, guaranteed
> I don't think this will actually be the case. You still have system level dependencies to worry about, which will differ on different OS' and distributions. (This is one of the big things that Docker solves).

Trueee, even if we have CI that checks that our package works on (e.g.: `{windows, mac, linux}`), there will always be a developer whose machine that falls outside of our tested configurations. The theory with this approach is that edge cases won't be common, and the majority of developers should be able to successfully work on the project since the primary OS configurations are tested

rfcs/0022-standard-tooling.md
rail: The idea of hash is that this is a unique docker tag that never changes. The hash can contain human readable strings (version, for example). We can probably use "unique name" instead of "hash"?

rfcs/0022-standard-tooling.md
rail: You can use your IDE, but your are on your own in this case. It'd be hard if we start supporting IDE-based (which one) development in parallel with the proposed workflow.

mitchhentges: > You can use your IDE, but your are on your own in this case.

Ok, it sounds like the intent is that we wouldn't merge platform-specific fixes into the projects.

> It'd be hard if we start supporting IDE-based (which one) development

I'd argue that it's not about specifically choosing an IDE to support, but rather not locking ourselves into workflows that _hamper_ the usage of IDEs/tools.

mozbhearsum: I think we might not be talking about the same thing here...when I say "it's an anti-feature to support any platforms that we don't need to", I mean that we shouldn't spend non-trivial amounts of time to make things like scriptworkers, ship it, or Balrog run on OS X or Windows.

I don't mean to say that we shouldn't accept small, easy fixes that make local development easier.

More concretely, if we needed to adjust some path joining code to use `os.path.join` or look in a OS-specific location for a config file, I think that's OK. That's trivial to add, and doesn't add a maintenance burden.

If we needed to adjust business logic that shells out or runs subprocesses because it only works on Linux, I would not want to take a change like that.

Does that clarify things at all?

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+
+To meet this new standard we will need to change try & staging releases as follows:
+* Try releases will use Balrog & Ship It dev
+* Project branch releases (eg: maple) will use Balrog & Ship It stage
+
+## Notes on `mozilla-releng/services`
+
+`mozilla-releng/services` is where a large number of our applications live. They are currently built and managed with `Nix` and a special command called `please`, which wraps docker, tox, and other tools. Upon acceptance of this RFC these will be considered deprecated, and work towards removing them fully. As each project is converted, we will remove the `Nix` and `please` bits from it. When all projects have been converted, any remaining `Nix` and `please` references will be removed from the repository. (It is not practical to convert all projects at once, so this piecemeal method is necessary.)
+
+This RFC does not necessitate changing the overall structure of this repo. Shared libraries may continue to live at the toplevel `lib` directory, and be used by the projects in `src`. In order to accommodate this, Docker builds will need to be run from the root of the repository. For example, here is a potential Dockerfile for `shipit/api`:
```
garbas: Structure of `mozilla-releng/services` was built with Nix in mind. Mono repos with current tools selected will be a pain to maintain and we will end up again building the framework. I would split up all the projects out from there as we migrate to this RFC.

rfcs/0022-standard-tooling.md
srfraser: Does this mean that github's usual pull request process will put things directly into staging? Will we have to remember to compare and PR into `dev` instead of `master`? Or would it make sense for the default to be the least risky, and have PRs get merged into `dev` (which is `master`) and then have `staging` and `production` branch off there?

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+
+# Details
+
+## Development
+
+Docker and Docker Compose will be used to build, run, and test applications locally. This ensures that all developers, CI, and deployed environments are as identical as is reasonable. Using Docker means that we will explicitly not be pinning system level packages anymore. This is an explicit trade-off that we are making -- the cost/complexity of doing so is not worth the small benefits we get from it.
+
+Python applications should use `python:3-slim` as their base image for production and `python:3` as their base image for local development. Javascript applications should use `node:current` as their base image.
```
srfraser: Presumably tests should be run in python:3-slim, though?

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Dependency Management
+
+Python projects will specify their dependencies in three files (as is standard for `pip-compile-multi`):
+* `requirements/base.in` will contain direct dependencies required for running the application
+* `requirements/test.in` will contain direct dependencies required for running tests
+* `requirements/local.in` will contain `tox`
+
+When any of these files change, `pip-compile-multi` will be used to generate the full list of pinned dependencies. Pyup.io will be used to keep them up-to-date.
```
srfraser: Could we specify that pyup.io, if configured to put in PRs, should do so all at once? ?Having multiple conflicting PRs to resolve is annoying (For example, since pyup.io doesn't know that boto3 and botocore are codependent it puts in PRs for each individually, which don't pass tests)

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+
+Javascript projects will specify their dependencies in `package.json`. As is the usual for Javascript projects, direct dependencies will be listed in `dependencies` and direct test and local development dependencies will be listed in `devDependencies`. When either of this lists change, `yarn lock` will be used to generate the full list of pinned dependencies. Greenkeeper.io will be used to keep them up-to-date.
+
+## Testing
+
+### Unit Tests
+
+All testing will be done within Docker container.
+
+Python projects will wrap Docker with `tox` while Javascript projects will wrap Docker with `yarn`. These wrappers will build the necessary Docker image, and then run the tests within it. This is to maintain a familiar, well-known entrypoint for running tests, while ensuring they run in a consistent way for all developers.
```
srfraser: Since it's bitten me recently, could we have a note about what sort of test failures should block merging? The situation I had was that I had removed some obsolete, but tested, code. This reduced the covered percentage of code, and so merging was blocked. If we include some guidance about things we've learned to do and not do here, that would be helpful.

srfraser:
If it's possible, I'd like some extra information about how to determine if the new structure is working as we'd hoped. Perhaps connected to the 'Motivation' section, but explicitly listing the goals of the new structure, how we would tell if the new way is helping, what sort of things we would be looking for in a new tool or process that someone would like to suggest we use instead.

rfcs/0022-standard-tooling.md
mitchhentges: Absolutely, that makes sense, thanks for clarifying :)

As a side-note, a criteria that I'd still like recommended in our new standard is documentation to set up each $project's environment manually. Sometimes, projects might lean on the existence of a docker image, then avoid writing manual-setup-documentation since "you can just use the docker image".
A nice balance I've seen is project documentation starting by explaining how to use the docker image, but they also have a section on creating the environment on their native platform (e.g.: Linux). What do you think?

rfcs/0022-standard-tooling.md
mitchhentges: That's a really good point!
To be verbose, IMO there's three options here:

1. (causes the issue Simon ran into) Don't allow merging a PR if it lowers the overall coverage %. Downside: PRs that remove code lowers the coverage %, but don't add new uncovered code, still get blocked from merging
2. Don't allow merging a PR if it **both** lowers the overall coverage % **and** adds new uncovered code
3. Don't block PRs based on code coverage. This approach acknowledges that some code is sufficiently simple that it doesn't need to be tested, but rather leans on code review and staging environments to realize when additional coverage is required.

My guess is that option 2 fits our team the best and acknowledges the edge that Simon ran into. However, if there's any hope of option 3, I'll gladly argue for it :smile:

rfcs/0022-standard-tooling.md
mozbhearsum: I think it's a fantastic idea to have documentation :). Dockerfiles are not docs!

rfcs/0022-standard-tooling.md
mozbhearsum: Hm, OK. I'm not as familiar with scriptworker deployments. Is this because there is nothing to be gained from spending time in stage? Eg: we've already tested everything in dev?

rfcs/0022-standard-tooling.md
mozbhearsum: Do you have any suggestions on what we'd do with `lib` (in particular, the shared code for auth and other such things) if we split things up? I have no strong feelings about the mono repo vs. multiple repos, but I'm not sure where that shared code would live in a multiple repo world.

rfcs/0022-standard-tooling.md
mitchhentges: It's more that there's only one user of each script for each environment.
If I'm going to live-test a change to `signingscript`, I'm going to do it off of my branch (so that if there's issues, I don't need to re-PR). Otherwise, if I'm not testing in `dev`, then I'm probably not going to do a `staging` release.

This is because we don't have a "pool" of early-adopters that can check our work - we just have a "production" worker, a "dep" worker, and a "dev" worker. "production" and "dep" are both expected to not fail due to internal issues, and "dev" isn't used regularly in a way that it'll catch failures like a "staging" env.

rfcs/0022-standard-tooling.md
mozbhearsum: You're correct - under this proposal merged PRs would directly to `stage`. The idea here is that `dev` is a deployed environment that can be used *prior* to having a merged PR (which I believe is the only realistic way to test some patches -- cc @rail). In my head, it's basically an ad-hoc testing environment. You're not necessarily going to use it for every patch, just patches you need to test in a deployed environment prior to merging.

If that's not an important thing to maintain, I fully agree that having the `dev` environment track the `master` branch makes sense.

mozbhearsum: There's also another thread here about having `staging` and `production` branches in the repo vs. just deploying specific builds off of master (the latter is what this RFC assumes at the moment). That's something worth discussing more, too.

rfcs/0022-standard-tooling.md
mozbhearsum: Ideally, yes, but it's actually tricky to do so at the moment, as the slim images do not install `git`. So you either need to create your own image based on it, or use the larger image. Perhaps we could maintain a our own derivative of python:3-slim that installs the couple extra tools we need?

rfcs/0022-standard-tooling.md
mozbhearsum: Yes, that's a great idea!

rfcs/0022-standard-tooling.md
mozbhearsum: I think this is fair to be in scope for this RFC. I'll get a strawman added to this RFC and we can try to come to consenus on it. I want to make sure @escapewindow gets a chance to weigh in, as I know he's got strong feelings about coverage.

rfcs/0022-standard-tooling.md
mozbhearsum: Hm. This might be scope creeping a bit, but I wonder if this is a good time to consider having 3 proper environments for scriptworkers (arguably, it's required as this RFC is currently written....).

mitchhentges: That sounds valuable, but we'll need to change how we use our scriptworkers to take advantage of our new environments if we decide to do so. I'm not having any workflows jump to mind :thinking:

rfcs/0022-standard-tooling.md
La0: You _may_ want to wait a bit for implementing a decision task engine, and rely on the one from mozilla-central, [once it's setup as a standalone library](https://hg.mozilla.org/ci/taskgraph/).

mitchhentges: Absolutely, we wouldn't want to write a decision task engine for the umpteenth time :laughing: I think we're aligned there. If we _did_ decide to do a fatter decision task rather than leaning on `.taskcluster.yml`, I'm guessing that we'd wait for `taskgraph` or an alternative to be ready

La0:
I'm working with @marco-c on a small project named [task-boot](https://github.com/mozilla/task-boot/) that may be releavant to your interests. It aims to implement CI/CD workflows in Taskcluster around Docker images.

Right now it's able to:
* clone a git repo
* build a docker image in a Taskcluster task without `dind`, by using the excellent project [img](https://github.com/genuinetools/img/)
* push that docker image to a Docker repo, reading credentials from a Taskcluster secret
* build multiple docker images using a `docker-compose.yml` file
* build/update a Taskcluster hook
* write docker images as Taskcluster artifacts
* use those artifacts in another task to push them (allow for workflows like: build on pull-request and build/push on tag/prod branch)

The goal is to keep the smallest code footprint possible, mostly producing glue code between tools & Taskcluster APIs.

TaskBoot is used by [bugbug](https://github.com/mozilla/bugbug/) to produce Docker images on pull requests and branch pushes, pushing them only when a tag is created. Bugbug has 7 images bases on a _fat_ base image

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+Docker and Docker Compose will be used to build, run, and test applications locally. This ensures that all developers, CI, and deployed environments are as identical as is reasonable. Using Docker means that we will explicitly not be pinning system level packages anymore. This is an explicit trade-off that we are making -- the cost/complexity of doing so is not worth the small benefits we get from it.
+
+Python applications should use `python:3-slim` as their base image for production and `python:3` as their base image for local development. Javascript applications should use `node:current` as their base image.
+
+## Dependency Management
+
+Python projects will specify their dependencies in three files (as is standard for `pip-compile-multi`):
+* `requirements/base.in` will contain direct dependencies required for running the application
+* `requirements/test.in` will contain direct dependencies required for running tests
+* `requirements/local.in` will contain `tox`
```
tomprince: I'd call this dev, and include things like `pip-compile-multi` used for development but not testing.

rfcs/0022-standard-tooling.md
tomprince: I think the deployment story here is for server applications and not libraries or developer tools. I think we should consider publishing libraries (scriptworker itself is currently a library, `mozllla-version` is definitely a library, `mardor` id a developer tool) to pypi, but we should not publish server applications to pypi.

rfcs/0022-standard-tooling.md
tomprince: I think we should aim to have easy to create environments, rather than have a fixed number of shared testing environments (see [here](https://alexgaynor.net/2016/jan/19/dont-have-environments/) for a more thorough discussion.

rfcs/0022-standard-tooling.md
mozbhearsum: I'm open to using `taskgraph` if it works well for smaller projects. In its current form, it may be far too heavyweight for our purposes, but maybe it looks/feels different once its out of the tree?

rfcs/0022-standard-tooling.md
garbas: as we do with all other projects. "reimplement" (copy-paste in this case) in project repo. those projects in lib were anyway on the path to be split.

rfcs/0022-standard-tooling.md
mozbhearsum: Yeah, I'm not sure either. This is a bit of a tangent, but it may be worth making a list of what we have now (both scriptworkers and other systems), what they're used for, and see how they do or do not align.

rfcs/0022-standard-tooling.md
catlee: just to chime in on the multi-platform issue...I think it makes sense to make some minor concessions to make development possible or easier on different platforms than what we are deploying to. However, I don't think it makes any sense to spend any time making the application code itself work across multiple platforms. The majority of our code will only ever run in production on linux, let's not waste time trying to make it work on non-linux platforms.

rfcs/0022-standard-tooling.md
catlee: If we do this, should parts of `lib` be packaged as distinct python modules?

rfcs/0022-standard-tooling.md
catlee: Do we gain anything by depending on slim instead of the regular image? https://docs.docker.com/samples/library/python/ recommends against using the slim images for general use:
```
Unless you are working in an environment where only the python image will be deployed and you have space constraints, we highly recommend using the default image of this repository. 
```

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+
+COPY ${project_dir}/setup.py /app
+COPY lib/ /app/lib/
+```
+
+And from the root of the `services` repo, we can use `docker build -t shipit_api -f src/shipit/api/Dockerfile .` to build it.
+
+# Open Questions
+
+* Does Taskcluster's version of Docker/docker-in-docker support multistage builds?
```
JohanLorenzo: It doesn't yet. Multistage builds were added in Docker 17.05 [1], which is API 1.29 [2]. Taskcluster uses API 1.18 [3]. We need [bug 1501720](https://bugzilla.mozilla.org/show_bug.cgi?id=1501720) to be deployed once more. 

[1] https://docs.docker.com/develop/develop-images/multistage-build/
[2] https://docs.docker.com/develop/sdk/#docker-ee-and-ce-api-mismatch
[3] https://bugzilla.mozilla.org/show_bug.cgi?id=1513947#c0

rfcs/0022-standard-tooling.md
JohanLorenzo: If we call it `dev`, I'm afraid people may confuse it with the `dev` environment. I would suggest to make these 2 terms not overlap and rename one or the other.

rfcs/0022-standard-tooling.md
JohanLorenzo: I'd lean towards a decision task too. I've seen projects outside of releng (like deepspeech or servo) starting with everything in `.taskcluster.yml` and then needing a decision task because `.taskcluster.yml` was too hard to maintain. It's also easier to run a simple script locally, than to setup json-e parser to ensure `.taskcluster.yml` is well defined. 

Furthermore, having decision tasks enables `.taskcluster.yml` to be factorized within. For instance, common metadata in android-components is shared at:  https://github.com/mozilla-mobile/android-components/blob/d98b984bfde5fbe62c921347bdb92b3c28b09977/.taskcluster.yml#L78.  Moreover, if we go the decision task way, we may be able to have `.taskcluster.yml` pretty similar from one repo to another. That's what we currently have in android-components, fenix and reference-browser. 

That said, I agree `taskgraph` is not ready yet, and we may want a simpler library. @mitchhentges started `decisionlib`, maybe that's the direction we want to head to?

rfcs/0022-standard-tooling.md
mozbhearsum: I somehow missed @tomprince's comment yesterday...

I totally agree that having easy to create environments is a good idea. How practical do you think this is when we have external dependencies to systems we don't control? Even things such as Balrog and Ship It are not really in our control at this point, as CloudOps owns the deployments.

rfcs/0022-standard-tooling.md
mozbhearsum: I'd be :+1: on that. Especially when it comes to the auth code (which can be tricky to get right), I hope we can avoid so many instances of it all over the place.

rfcs/0022-standard-tooling.md
mozbhearsum: Interesting! The main reason we're using `slim` is because CloudOps advised it when we were originally deploying Balrog there.

@oremj or @jbuck - how much are you concerned with Docker image size? Would it be problematic if we used the non-slim versions of Python images for Balrog (and future projects)? It looks like 3.7-slim is currently ~150MB, while 3.7 is ~1GB.

rfcs/0022-standard-tooling.md
mozbhearsum: Hmm, I thought @rail said he made it work...I could be wrong though.

rfcs/0022-standard-tooling.md
marco-c: Using the slim image might speed-up building/exporting the Docker images on Taskcluster (because of reduced I/O).

rfcs/0022-standard-tooling.md
marco-c: We were thinking about this for taskboot, but we decided to go with .taskcluster.yml until the taskgraph library is available. We wanted to avoid as much as possible to build additional things on top of the things that most Mozilla developers would already be accustomed to.

rfcs/0022-standard-tooling.md
jbuck: Not too concerned. It affects us more on AWS because each instance needs to pull down the docker image on boot and that can affect startup time. On GKE, each node only needs to fetch the image once so that pull time is amortized over many container starts

rfcs/0022-standard-tooling.md
tomprince: I've got [this](https://github.com/mozilla-releng/redo/pull/60/files) as a WIP example of using taskgraph on github. It could do with some simplifications, and depends on a couple of hacks currently, but it is a reasonable base.

rfcs/0022-standard-tooling.md
tomprince: I know at least the taskcluster team (and I think maybe other mozilla teams as well ([treeherder](https://github.com/mozilla/treeherder/blob/master/.yarnrc)) are using yarn. In this case, I think the mozilla ecosystem probably has slightly more weight than the wider one.

rfcs/0022-standard-tooling.md
mozbhearsum: Rail and I were talking about this a bit more, and after this long winding discussion we're starting to wonder whether it's too early to standardize on a specific Taskcluster task graph tool. There's at least 3 or 4 options by my count, and none of them feel particularly mature.

What do people think about just specifying "Taskcluster" for now, and we can come back to this at a later time, when these tools have been shaken out a bit more?

rfcs/0022-standard-tooling.md
mozbhearsum: Rail and I were talking about this thread a bit and came to the realization that Docker has more carrots to offer for some projects than others. For example, if you're running a webapp, you probably need a database and to bring up a frontend and a backend. It's way easier to do this with Docker than on your host system. If you're just running a scriptworker locally, the benefits aren't quite as obvious. However, it's worth pointing out that instead of running "docker-compose up", you're instead running "pip install -r", possibly creating a config, and then running a command. 

With that said, I still think it's important and valuable to have one officially supported way of running an app, and I think that Docker is it, especially because once scriptworkers have moved to GCP, the vast majority of our infrastructure will be deployed through Docker. This doesn't mean you can't run things on your host machine, but as @catlee said, we shouldn't be spending time making it work across multiple platforms/distros.

I don't mean to shut down discussion on this point, but I see a lot of benefits to making Docker the supported way, and only one drawback (people have to install Docker).

rfcs/0022-standard-tooling.md
mozbhearsum: In rereading this thread I got to wondering: do we acutally _use_ any of our `dev` environments? I know we have `dev` scriptworkers, but I'd argue those are actually `stage` scriptworkers. I know we don't use Balrog or Ship It dev.

If we don't actually have a use for `dev` environments, maybe we should stop creating/kill them? It might encourage us have more easily spun up testing environments, instead, like @tomprince wants.

mozbhearsum: And on the bake time for `stage`, I'm happy to drop that. That's something we cargo culted from the existing `services` repo release process, and I think it's clear that it doesn't make sense across the board (especially if we kill the `services` monorepo, and deploy those applications separately).

rfcs/0022-standard-tooling.md
mozbhearsum: Please correct me if I'm wrong, but it sounds like nobody feels strongly that we should keep `services` as a monorepo. It sounds like we should consider:
* Moving the `lib` code into proper shared libraries
* Splitting up the repo by project. There may be value in keeping some projects together (eg: the Ship It frontend and backend probably belong together, maybe some of the `releng` family should stick together too?).

rfcs/0022-standard-tooling.md
mitchhentges: Keep in mind that the only drawback isn't that people have to install docker: we also incur the complexity and security costs of having docker as part of our official workflow.

I agree that it makes sense to provide an easy and standardized way of testing our tools, so :+1:

rfcs/0022-standard-tooling.md
garbas: @catlee i'm -0 on creating more python modules. there might be few cases where that makes sense, but there is a cost of having more python modules and trying to keep everything in sync with multiple repositories is pain with any python package manager.

@mozbhearsum the auth part is actually quite easy, either with taskcluster and/or with auth0, I'd just copy paste those 20 lines from somewhere. I find it more error prone and (maintainance) cost to high to justify maintaining all those extra python modules and making sure we don't break anything when updating.

I find copy paste approach to beat the python tooling in most cases, especially looking in the long run.

rfcs/0022-standard-tooling.md
mozbhearsum: Having Docker part of the official workflow is a sunk cost, isn't it? Production is done with Docker containers (even for scriptworkers in the near future). Or do you separate production-with-Docker from development-with-Docker as a cost?

mitchhentges: Ah, yes, when docker is already used for production, adding it to dev just incurs the "people have to install docker" cost

rfcs/0022-standard-tooling.md
escapewindow: +1, it's early, and at some point we'll want to use taskgraph for anything that generates sensitive artifacts (docker images, python packages, etc.) because it will give us the ability to verify the graph via cot.

rfcs/0022-standard-tooling.md
escapewindow: I think we can make rules based on the repo. If we want a releng-wide standard, or just a standard for the scriptworker repos, I would suggest [100% line coverage](https://escapewindow.dreamwidth.org/247980.html) is a good start. Adding integration tests and additional functional/fuzzing tests beyond just line coverage is even better.

I do think that good [test docstrings](https://jml.io/pages/test-docstrings.html) are useful, and should be a requirement for any new tests. We can go back and add them to old tests at some point, if it makes sense to do so.

> I want to make sure @escapewindow gets a chance to weigh in, as I know he's got strong feelings about coverage.

Thanks, I do :)

I did see some complaints about 100% coverage not catching bustage before testing happened in staging. This is true for all unit tests: they only test your assumptions, and only so much as your tests cover. Possible solutions:

- double check your assumptions
- write functional tests to make sure your code matches your assumptions
- it's possible to test in staging before 100% code coverage, to make sure your assumptions are correct. Once things are working, you can add 100% coverage before review/merging.
- add integration tests to add extra assurances that your code works in a non-mocked environment

> PRs that remove code lowers the coverage %, but don't add new uncovered code, still get blocked from merging

Not a problem with 100% coverage. Removing code will still result in 100% coverage.

> Don't allow merging a PR if it both lowers the overall coverage % and adds new uncovered code

You have to actually look at, and reason about, the code coverage results to know this. With 100% coverage, you have a green or red status, and you can have Github block merging automatically.

> Don't block PRs based on code coverage.

[The new contributor you are helping may be you in six months.](https://escapewindow.dreamwidth.org/247980.html#future_changes)

rfcs/0022-standard-tooling.md
escapewindow: I could see having a recommended baking period for non-scriptworkers. We may need to see how/if we use a staging scriptworker env before we add that requirement to them.

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+
+The stage environment should be used as a final testing ground before pushing to production. Code typically bakes here for ~24h before moving to production.
+
+### Dev
+
+Dev builds will happen in response to push events on the `dev` branch. Once a build is completed, it will automatically deploy to the `dev` environment. For Python projects, this means pushing a `${project}_dev` tag to Dockerhub. For Javascript projects, this means publishing to S3.
+
+Anyone may force push to this branch to deploy their code. However, because we only have one dev environment and many developers, it should *not* be used it for any testing that can be completed with a local Docker container. This environment is typically reserved for testing that can only be accomplished in a deployed enviroment. Examples include:
+* Testing changes to Sentry or other CloudOps integrations
+* Testing changes that depend on services you cannot run locally (eg: hg.mozilla.org, Taskcluster)
```
escapewindow: The dev/staging branches seem to be more important for services that need to be deployed somewhere. For flexible pools where we can either run a docker instance locally that picks up Try jobs, or if we're able to say "please spin up `n` dev docker images of workerType `w` using docker image `i`", we could generate the docker images off any PR or fork.

rfcs/0022-standard-tooling.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Dependencies between applications
+
+Any application that depends on another application *must* use the same environment. Eg: ship it stage must use balrog stage, it cannot use balrog dev.
+
+Dependencies on external applications are a little more ambiguous, as not all external dependencies have a matching set of environments. Use your best judgement here. Eg: don't have production services depend on external staging services or vice versa.
+
+To meet this new standard we will need to change try & staging releases as follows:
+* Try releases will use Balrog & Ship It dev
+* Project branch releases (eg: maple) will use Balrog & Ship It stage
```
escapewindow: +1 to judgment on external deps. We may point at read-only production services from dev/staging envs in staging releases.

rfcs/0022-standard-tooling.md
escapewindow: It's possible to have [multiple python packages in a monorepo](https://github.com/escapewindow/scriptworker-scripts).

rfcs/0022-standard-tooling.md
escapewindow: Ah, also iscript scriptworkers will be running on mac hardware, not docker.

escapewindow:
Ben, Rail thanks for writing up this rfc. i understand the need for it and agree with it overall... I just wanted to clarify some points if we intend to include scriptworkers in the blanket term "releng apps". And weigh in with my 2c on coverage :)

rfcs/0022-standard-tooling.md
mitchhentges: I agree that spending time on testing is a good investment in the long-term, and there's benefits to having 100% coverage (like no longer being bite-able by removed-code-blocking-PRs).

However, I think it's important to acknowledge the cost and diminishing returns of tests.
For example, look at all these [simple getters in `scriptworker](https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/task.py#L117-L129). How much value is _really_ being added by adding tests?

Related: I come from statically-typed languages like Java and ~Rust, so I'm used to leaning on the compiler, and only adding tests once a component becomes sufficiently tough to reason about. Python has way less guarantees (even with `mypy`), so I can understand if the push for increased coverage is due to the dynamic nature of our primary language.

rfcs/0022-standard-tooling.md
mozbhearsum: Good points. I'll add something to the RFC to cover them.

rfcs/0022-standard-tooling.md
mozbhearsum: Thanks for all of the details here. A few things to follow-up on:
> If we want to include them under the umbrella term Releng applications

They are explicitly in scope here, for the record.

> change the scriptworker pools to exactly match the dev/staging/prod environment split (which would move dep signing back into the same pool as release/nightly, which isn't necessarily the best security-wise), or

> Only some scriptworkers have a dedicated dev pool.

Is there a reason we don't have dedicated dev pools for some scriptworkers, but we do for others?

I suspect I'm missing a subtle detail here, but I'm not sure why this would be so troublesome? I definitely understand the need to keep regular signing scriptworkers and dep signing separate, and I don't want to change that, but I think of these as parallel production environments, in the same way that we have production thunderbird signing scriptworkers. Dev is separate, and we could potentially even have dep and non-dep dev deployments if we wanted.

There still needs to be some more discussion and follow-up on the prod/stage/dev thing....I think it's pretty clear there's not a simple answer. I'll come back to that soon.

rfcs/0022-standard-tooling.md
escapewindow: There are diminishing returns to a fully mocked main() test, or simple getters. The burden of writing the test for the getter is very low, and if it's not worth writing a specific test for it, you get coverage by testing functions that call the getter. And again, the difference is a one-time hit to the person writing the code, or an ongoing burden on committers and reviewers to look at the coverage results to make sure that no new code is being added without coverage.

tomprince: The requirement for 100% test coverage is something that I've had a few discussions with @aki about.  While I think 100% coverage is valuable, I also think there are tests that provide negative value, and that requiring 100% can encourage writing those tests. And having worked on codebases with low-value tests, I come down strongly on the side of not having them, even if that means not having 100% coverage.

escapewindow: Which tests are you referring to? And how are they negative value? Depending on the answers, we might be able to discourage those types of tests.

rfcs/0022-standard-tooling.md
escapewindow: > Thanks for all of the details here. A few things to follow-up on:
> 
> > If we want to include them under the umbrella term Releng applications
> 
> They are explicitly in scope here, for the record.

Good to know, thank you.

> > change the scriptworker pools to exactly match the dev/staging/prod environment split (which would move dep signing back into the same pool as release/nightly, which isn't necessarily the best security-wise), or
> 
> > Only some scriptworkers have a dedicated dev pool.
> 
> Is there a reason we don't have dedicated dev pools for some scriptworkers, but we do for others?

It has to do with both growing organically with various people in charge of each scriptworker instance type, and the fact that it's not trivial to set up a new scriptworker in ec2+puppet. This should hopefully be a solved problem in the docker world.

> I suspect I'm missing a subtle detail here, but I'm not sure why this would be so troublesome? I definitely understand the need to keep regular signing scriptworkers and dep signing separate, and I don't want to change that, but I think of these as parallel production environments, in the same way that we have production thunderbird signing scriptworkers. Dev is separate, and we could potentially even have dep and non-dep dev deployments if we wanted.

I'm not sure what you mean by "this" in `why this would be so troublesome`. The RFC specifies how environments are rolled out, and which branches they roll out from; the specification matches how we roll out the current releng apps, but not how we currently roll out scriptworkers. Because there are some unknowns here about whether this model can work well for scriptworker deployments, I'm hesitant to sign off on explicit requirements that say we must follow that model. Perhaps we sign off on it with some unknowns, and revisit if/when there are problems. Perhaps we make the model a suggestion or a requirement-unless-compelling-reasons.

> There still needs to be some more discussion and follow-up on the prod/stage/dev thing....I think it's pretty clear there's not a simple answer. I'll come back to that soon.

+1

escapewindow: (Not sure we'll have a dedicated dev pool in the iscript world, though we could possibly keep partner-repack-1.mdc1 for that.)

rfcs/0022-standard-tooling.md
mitchhentges: > And again, the difference is a one-time hit to the person writing the code, or an ongoing burden on committers and reviewers to look at the coverage results to make sure that no new code is being added without coverage.

As an aside: I would argue that maintaining tests is not a one-time cost - requirements and functionality will change, and tests will need to be updated accordingly.

rfcs/0022-standard-tooling.md
nthomas-mozilla: In general these rules make sense, but I'm not keen on the distinction between try and project branches and think they should both default to stage. Are there some privilege or scopes arguments for the system you suggest ?

My reasoning is that releases on try are unlikely to need to test Balrog changes, plus dev will probably end up stale (either on a PR or some old master rev) and that will lead to false negatives or positives. For those times we do want Balrog dev we could have another arg for the [release selector](https://firefox-source-docs.mozilla.org/tools/try/selectors/release.html) to make that change; this be extended to other services too.

rfcs/0022-standard-tooling.md
mozbhearsum: > I'm not sure what you mean by "this" in why this would be so troublesome. The RFC specifies how environments are rolled out, and which branches they roll out from; the specification matches how we roll out the current releng apps, but not how we currently roll out scriptworkers. Because there are some unknowns here about whether this model can work well for scriptworker deployments, I'm hesitant to sign off on explicit requirements that say we must follow that model. Perhaps we sign off on it with some unknowns, and revisit if/when there are problems. Perhaps we make the model a suggestion or a requirement-unless-compelling-reasons.

"this" means scriptworkers following the same environment/deployment models as other apps - specifically, a dedicated non-production pool (which I think we'll end up calling stage in the end), and possibly ad-hoc development environments.

I understand that it's different right now, but I'm not entirely clear on why it should/needs to be. You mentioned that it's not trivial to set-up new dev enviroments in the non-Docker world, are there things other than that, too? Or are we mostly talking about unknown unknowns?

rfcs/0022-standard-tooling.md
mozbhearsum: It's a good point that things that operate on a pull model (such a scriptworkers) don't benefit as much from dedicated dev/stage environments. I think this is definitely true for early testing ("does my code work at all"), but I still think there's value in having a permanent, deployed environment. Even though things in this category don't have external systems talking to them directly, they are still affected by and affect other things. Eg: if a scriptworker is producing bad artifacts or other things consumed by downstream systems, it may not be caught by the sort of isolated testing that I think you're talking about here (correct me if I'm wrong, though).

rfcs/0022-standard-tooling.md
mozbhearsum: This is a good point about dev getting out of date. We largely proposed try = dev because of commit levels, and try being the closed equivalent to a "dev" tree that we have.

In a world where try releases do or can use a separate Balrog instance, I could see them being used more for testing Balrog changes (in fact, I've already started doing that for the Google Cloud changes, now that './mach try release' is pretty trivial to use) -- but I think you're right the usual case probably wants something close to the production Balrog code.

There's been a lot in the comments already about the purpose, function, and necessity of the dev & stage enviroments. I think this part of the RFC is going to change a lot soon, and I'll take this into account.

> My reasoning is that releases on try are unlikely to need to test Balrog changes, plus dev will probably end up stale (either on a PR or some old master rev) and that will lead to false negatives or positives. For those times we do want Balrog dev we could have another arg for the release selector to make that change; this be extended to other services too.

This is a really nice idea. I think it would work well for Balrog and other webapps. Scriptworkers might be a little tougher (I think you'd need more changes to the config/tree to make it work), but maybe still doable.

rfcs/0022-standard-tooling.md
mozbhearsum: I think we've more or less reached consensus here, so I'm resolving this. Please re-open or add a new comment if you disagree!

rfcs/0022-standard-tooling.md
mozbhearsum: Looks like I missed this comment, sorry for the slow reply.

The reason that `local.in` is meant to include Tox is because we need a way to Bootstrap the in-Docker tests. The idea is that you run `tox`, which runs Docker, and then the test dependencies from `test.in` (pytest, et al) are installed inside of the container. Arguably, it would be OK to rely on `tox` being available instead of requiring a specific virtualenv for it. The same thing could apply for `pip-compile-multi`. I'm not sure if either needs a virtualenv during local development upon further reflection.

However, I think we do need tox in a requirements file for the purpose of generating a Docker image to run tests inside for CI (I don't think we should just "pip install" it there).

mozbhearsum: > If we call it dev, I'm afraid people may confuse it with the dev environment. I would suggest to make these 2 terms not overlap and rename one or the other.

If we continue with deployed dev environments, I agree we should avoid this. If we don't have those I'm less concerned.

rfcs/0022-standard-tooling.md
escapewindow: > It's a good point that things that operate on a pull model (such a scriptworkers) don't benefit as much from dedicated dev/stage environments. I think this is definitely true for early testing ("does my code work at all"), but I still think there's value in having a permanent, deployed environment. Even though things in this category don't have external systems talking to them directly, they are still affected by and affect other things. Eg: if a scriptworker is producing bad artifacts or other things consumed by downstream systems, it may not be caught by the sort of isolated testing that I think you're talking about here (correct me if I'm wrong, though).

This is only true if something is feeding them work. There are several ways this can happen:

1. The staging scriptworkers are the same workerType (pool) as, say depsigning or prod signing. This means they're production, not staging; they will claim production tasks and turn the tree green or red. This is possible, but could lead to more tree bustage than otherwise. And it's not really staging.

2. We have a separate set of workerTypes for staging. However, we need to feed them tasks.

One way to do this would be for all production graphs to fire a set of "staging" tasks that also get fired, but are hidden from Treeherder (tier3?). At one extreme, this can mean doubling our load, if we duplicate the whole graph but point the scriptworker tasks at staging. At the other extreme, we create a bare bones build->signing->beetmover etc tasks. We have to make sure that no taskcluster/ changes in-tree break these staging tasks; other people will be ignoring these task statuses.

Another way would be for us to fire off a try push whenever we want to exercise staging scriptworkers. We could possibly have a flag to switch to our staging scriptworker workerTypes automatically, saving us configuration time and overhead. Unless we automate this, this will be a manual test when we want to test something specific in staging.

An important thing to note: staging and dev environments are not as useful in certain scriptworkers. There is no staging or dev Google Play server/account, or Snap, or a.m.o for langpack signing, so the value of running a staging task through them is minimal. Our staging balrog, bouncer, and ship-it services, as well as our staging s3 buckets, are very dependent on existing release history state: if we run a staging release against unknown partials or reusing a concurrent release version, we'll see bustage in staging that would not exist in production.

Because of this, I'm meh on the idea of a staging env for scriptworkers in general, and more interested in how to make it easy to spin up dev environments and reset our staging release upstream services so we can reliably do a staging release with full partials and expect it to go green.

escapewindow: (This is just gecko, as well; maintaining these separate task graphs for thunderbird and all of our mobile and services github repos will be additional work.)

One thing we could do instead of staging (or in addition), is to keep bumping the workerTypes, e.g. signing-linux-v99. This would allow for using the new workerTypes in a PR or commit, and pinpointing which commit broke things should it break. This may be good for a general rollout, but if we need to replace all running workerTypes for all historical commits, we'd have to replace which docker images the various workerTypes point at to a newer one.

escapewindow: More brainstorming:

We could have a custom action that takes an existing nightly or release graph and replaces the scriptworker tasks with new scriptworker tasks run in staging. This would be a shortcut to running the whole graph in staging, and could be useful, with the caveats

- these will rely on internal testing only, e.g. the signing task verifying the mar signature. Since there are no downstream tasks like tests or repackage running in this partial staging graph, those tasks won't verify the outputs
- these will download the production release or nightly artifacts, and then talk to staging services. We may have a dependency issue, e.g. where the balrog tasks run in the wrong order (push live before populating the release info in balrog-stage? bouncer alias before populating the release?) unless we add the upstream scriptworker deps appropriately
- we still have the issue of needing to set up staging services properly before this will work.

A lot of my thoughts are revolving around this last bit. I think I would be more enthusiastic about a scriptworker staging environment if we could automatically run things against our staging services and have them be reliably green, without having to dig around for good partials or release versions that haven't pushed to s3 yet. Perhaps rail's service reset button idea would help us here.

mozbhearsum: > Because of this, I'm meh on the idea of a staging env for scriptworkers in general, and more interested in how to make it easy to spin up dev environments and reset our staging release upstream services so we can reliably do a staging release with full partials and expect it to go green.

Are we getting hung up naming here? We already have non-production scriptworkers that interact with non-production upstream services like Balrog, Bouncer, et al.

Or are you saying you want non-production environments that run code that hasn't been reviewed or landed yet?

escapewindow: It's not just the name, it's automatically spinning up scriptworker instances on master-merge that may not be used, or may be used to give broken or incomplete results. I think it would make sense to show that this environment serves a purpose and is useful and then implement it, rather than mandate it be implemented and then find a use for it.

rfcs/0022-standard-tooling.md
mozbhearsum: Given how complex of an issue this is, I'm going to declare it out of scope for now.

rfcs/0022-standard-tooling.md
mozbhearsum: You're right, it doesn't work!

mozbhearsum:
I just pushed an update that I _think_ resolves all of the outstanding comments.

PR #24, opened 2019-05-30
Adds simplify-cot RFC

mitchhentges:
@escapewindow, because you're the Chain of Trust king!
[rendered](https://github.com/mitchhentges/releng-rfcs/blob/simplify-cot/rfcs/0024-cot-reconstruction-only-on-decision-task.md)

### DIFF
--- a/rfcs/0024-cot-reconstruction-only-on-decision-task.md
+++ b/rfcs/0024-cot-reconstruction-only-on-decision-task.md
@@ -1,0 +1,67 @@
+# RFC 24 - Only Perform Chain of Trust Task-Reconstruction Once per Graph
+* Comments: [#24](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/24)
+* Proposed by: @mitchhentges
+
+# Summary
+
+Only perform "task recontruction from `.taskcluster.yml`" in the decision task, and append security
+information such as `is_commit_trusted` (is commit from the source repository in the primary branch)
+to the signed `chainOfTrust.json` file.
+Follow-up security-sensitive tasks such as for signing or pushing to Google Play will use this
+appended security information to make decisions, rather than having to rebuild the decision task from
+scratch, once for each task. 
+
+## Motivation
+
+**We can simplify projects that run `cot`-enabled tasks**
+
+Taskcluster allows scheduling tasks with default values. However, `cot` is not aware of these defaults.
+So, we have had to add unnecessary explicit default values to our created tasks. If we no longer
+do decision-task-rebuilding from follow-up tasks, we can avoid this.
+
+**We can simplify the `cot` work that happens after the decision task**
+
+By simplifying the work that `cot` does for follow-up tasks, we reduce bustage, 
+improve maintainability, and improve performance.
+ 
+
+# Details
+
+Before running security-sensitive builds (such as signing artifacts, or pushing to Google Play),
+we verify that the task group meets at least the following two criteria:
+* The entity that triggered the task group (user or scheduled hook) have sufficient Taskcluster scopes
+* The project revision being operated on exists in the primary repository (not a fork) on the primary
+branch (e.g.: `master`).
+
+The first criteria is checked by Taskcluster, but Chain of Trust verifies the second point by:
+1. For each `cot`-enabled task in the group:
+    1. Fetch the `MOBILE_HEAD_...` values from the decision task
+    2. Assert that the `MOBILE_HEAD_...` values are correct by rebuilding the decision task
+        1. Download `.taskcluster.yml` from the repository, branch and revision specified by `MOBILE_HEAD_...`
+        2. Rebuild a `jsone` context for the task group
+        3. Apply the `jsone` context to the `.taskcluster.yml` to get a decision task definition
+        4. Verify the created decision task definition against the real decision task that spawn the current task
+    3. Perform task-specific checks, such as verifying that the (now-trusted) `MOBILE_HEAD_...` values correspond to a
+    trusted source, e.g.: `mozilla-mobile/fenix`, `master` branch
+    4. ...
+
+This RFC recommends that we keep this procedure, but we only do it in the decision task.
+Then, as part of that decision task, we will put processed values (e.g.: `is_commit_trusted`) into `chainOfTrust.json`
+where it can be used by follow-up tasks. So:
+1. In decision task:
+    1. Fetch repository values, rebuild decision task from `.taskcluster.yml` as described above
+    2. Perform decision-task-specific checks
+    3. Encode properties like `is_commit_trusted` into `chainOfTrust.json`
+    4. Artifacts (including `chainOfTrust.json`) are signed with a trusted key
+2. In follow-up tasks:
+    1. Verify `chainOfTrust.json` is signed with the trusted key
+    2. Get properties from `chainOfTrust.json`, then perform task-specific checks
+    3. ...
+
+
+# Open Questions
+
+* Are there any negative security implications of this that have been missed? 
+
+# Implementation
+


rfcs/0024-cot-reconstruction-only-on-decision-task.md
```diff
@@ -0,0 +1,67 @@
...
+    4. Artifacts (including `chainOfTrust.json`) are signed with a trusted key
+2. In follow-up tasks:
+    1. Verify `chainOfTrust.json` is signed with the trusted key
+    2. Get properties from `chainOfTrust.json`, then perform task-specific checks
+    3. ...
+
+
+# Open Questions
+
+* Are there any negative security implications of this that have been missed?
```
escapewindow: Yes, it bypasses most security checks.

mitchhentges:
> Afaict, if someone runs an arbitrary decision task on a trusted decision worker, they can mark the commit as trusted and pwn the whole stack. How do we prevent that?

The task payload of the decision task doesn't have the `is_commit_trusted` value. Rather, when the worker creates `chain-of-trust.json` and verifies the decision task against `.taskcluster.yml`, _that's_ where `is_commit_trusted` is calculated.

tomprince:
Currently, neither docker-worker nor generic-worker do any verification of the task against anything, and I don't think they should, as doing that verification involves a bunch of policy and logic that aren't generic.

We could have a task that gets scheduled by the decision task, and runs on a scriptworker to verify the decision task, and then have other downstream scriptworker tasks depend on that and used its signed verification of the decision task as part of their chain-of-trust verification. This adds a fair bit of complexity to security sensitive code, though.

escapewindow:
+1 that neither docker-worker nor generic-worker currently do any verification. They may at some point, once cot verification is in a standalone tool, but until that's both security-signed-off and battle tested, I'd be wary of turning off full chain verification from the scriptworkers.

I also thought of the immediately-verify-the-decision-task-task solution, and have the same concerns: we would still have to be able to verify the whole chain at that task, which means scriptworker needs that functionality, and we add graph complexity to save a little bit of runtime in downstream scriptworker cot verification. We also previously ruled out having a scriptworker run all decision tasks, because that introduces significant friction in development.

mitchhentges:
> They may at some point, once cot verification is in a standalone tool, but until that's both security-signed-off and battle tested, I'd be wary of turning off full chain verification from the scriptworkers.

Interesting, are there currently other plans to split cot verification into its own component? Do we have a ticket for that already? Sounds like we'll really be splitting `scriptworker` into a bunch of pieces:
* Daemon
* Client
* Cot
* ... More?

-----

FWIW, what if we had a new worker type for decision tasks? It wouldn't be a `scriptworker`, but would rather be very similar to `docker-worker`/`generic-worker` - with decision-task-verification built-in.
What do you think?

escapewindow:
Yes, we're planning on potentially splitting scriptworker into 3 pieces.



> FWIW, what if we had a new worker type for decision tasks? It wouldn't be a `scriptworker`, but would rather be very similar to `docker-worker`/`generic-worker` - with decision-task-verification built-in.
> What do you think?

That's possible. However, the taskcluster team is currently close to finally reducing the number of official worker implementations from 3 to 1 (taskcluster-, docker-worker EOL, generic-worker runs all of those tasks) and this would add a new one. And to me, the math doesn't add up:

- we create and maintain a new worker implementation codebase. we have to keep this up to date
- we have to create and maintain new AMIs for this worker implementation
- the schema and features need to stay lock step with docker- or generic-worker, or we have to special case decision task definitions in taskgraph to avoid someone adding something incompatible to a decision task definition
- we may want to look even more carefully at decision worker provisioning security, depending
- we are blocked on moving cot verification out of scriptworker into a standalone piece, which we've talked about for a long time, but isn't started, let alone close to deployment
- we still have to keep `is_commit_trusted` logic in cot verification, for the decision task workers to be able to calculate if the commit is trusted, so we don't lose any complexity in cot.verify
- however, we do win by not having to run that verification on every scriptworker run. However, all of cot verification can be measured in a handful of seconds currently

Overall that sounds like significantly more time and effort spent to save a little bit of runtime cost on scriptworkers. Am I missing something?

rfcs/0024-cot-reconstruction-only-on-decision-task.md
```diff
@@ -0,0 +1,67 @@
...
+appended security information to make decisions, rather than having to rebuild the decision task from
+scratch, once for each task. 
+
+## Motivation
+
+**We can simplify projects that run `cot`-enabled tasks**
+
+Taskcluster allows scheduling tasks with default values. However, `cot` is not aware of these defaults.
+So, we have had to add unnecessary explicit default values to our created tasks. If we no longer
+do decision-task-rebuilding from follow-up tasks, we can avoid this.
```
mitchhentges: Looking at this more clearly afterwards, this point isn't true, since we still need default properties defined in the decision task so it can be re-created.

mitchhentges:
Hmmm, great points.
To clarify, I'm not too worried about the performance improvements of this, because I agree that that would be negligible :)

There's two reasons that I'm pushing to re-organize chain-of-trust validation:
1. It's currently hard to reason about CoT since it's spread over multiple workers. Understanding why my signing task failed because a _different_ task's payload had an untrusted repository was tough to grok the first time around
2. We have to wait until the `scriptworker` tasks run before we're notified of chain-of-trust failures. This usually means waiting for the long build to complete before a CoT-checking worker (such as `signingscript`) runs. If task groups "failed faster", then it would be easier to iterate on task groups that are failing chain-of-trust.

Looking at this closer with point 2 in mind, I wonder if this proposal would make more sense if it took one step further: have CoT verification for all tasks happen as part of the decision task (though artifact checksum verification would still happen in each task of course, since artifacts aren't known at decision-task-time). 
This solution addresses my two ideals above: it's easier to reason about since CoT happens all at once, and the decision task (the first task in the group) will fail if we don't pass CoT's requirements
This improves on my initial proposal by removing the need to push around temporary values like `is_commit_trusted`.

-----

So far, I've side-stepped your concern about how we'd manage the worker :thinking: :bulb:
I agree that forking `generic-worker` sounds like a significant cost.
Perhaps this logic would belong as part of `generic-worker`? Tom noted why this wouldn't make sense:

> Currently, neither docker-worker nor generic-worker do any verification of the task against anything, and I don't think they should, as doing that verification involves a bunch of policy and logic that aren't generic.

I don't fully understand, though: `generic-worker` already has CoT functionality (`features: chainOfTrust: true`). If this "decision-task-specific" CoT verification was toggled with its own `features` flag, then would it make sense to augment `generic-worker`?

-----

TL;DR, three options as I understand them:
1. Keep CoT as-is, with weaknesses in understanding CoT and delayed failures
2. Fork `generic-worker` into `decision-worker`, maintain AMI, keep compatible with `generic-worker` schema. Weakness: lot of work and effort, confusion for users when `decision-worker` task schema falls behind `generic-worker`
3. Add CoT logic to `generic-worker`. Weaknesses: `generic-worker` is now `less-generic-worker`

tomprince:
From the sounds of it, the main desire is for earlier feedback on Chain-of-Trust. There are a couple of things we can do that don't require any changes to the workers.
1. You can run the CoT verification code manually against any task with something like
```
verify_cot --cleanup --cot-product mobile --task-type decision <task-id>
```
as soon as the task is complete.
2. We could have the decision task run (some?) of the CoT verification on itself and/or the tasks it creates to fail early if it fails, but not try to re-use that downstream. There is some difficulty here in that CoT verification code is not in-tree and so finding the right version of the code to run would be tricky.

It sounds like the (1) address the concern about iterating on CoT tasks, which may be sufficent?

escapewindow:
> I don't fully understand, though: `generic-worker` already has CoT functionality (`features: chainOfTrust: true`

This generates the chain of trust artifacts for tasks run in the worker (`public/chain-of-trust.json` and `public/chain-of-trust.json.sig`). Without these, scriptworker can't verify the task and upstreams. And +1 to the `verify_cot` commandline tool helping debug issues. Once the upstream tasks have completed, you can even run `verify_cot` against, say, a build, before the build finishes. The decision or action tasks are the only ones that need to complete before you can verify them, since CoT verification downloads artifacts from them.

> it's easier to reason about since CoT happens all at once

We still have to verify CoT to make sure artifacts at rest haven't been modified since upload. So instead of all of CoT verification happening all at once in the scriptworker, it would happen twice, in the decision task and scriptworker.

mitchhentges:
I agree that the `verify_cot` CLI is useful, but I think that there'd still be two significant flaws:
1. It's still confusing for devs who aren't aware of CoT why a follow-up task (like signing) is failing due to the definition of the decision task.
2. In automation, if a task group has a CoT failure, it still waits until after the build finishes before CoT fails a build. Having to remember to run the `verify_cot` script adds manual work.

> We still have to verify CoT to make sure artifacts at rest haven't been modified since upload. So instead of all of CoT verification happening all at once in the scriptworker, it would happen twice, in the decision task and scriptworker.

True, but these are two different concerns. Verifying the integrity of the definitions of each task in the group (rebuild the decision task, check follow-up tasks, other CoT logic) is significantly different from verifying that an artifact is passed from task A to task B without being tampered. I don't believe that these two unrelated purposes need to be handled at the same time

escapewindow:
1. any changes like this have so many prerequisites that even if we agreed this was the right approach, by the time we can implement this successfully we may want to revisit what cot looks like in general. There have been rumblings that cot should be more a part of the taskcluster platform, but what that looks like is still up in the air. Plus the taskcluster team hasn't yet fully signed off on that yet. There are enough variables here that I think we can consider that you've proposed this, and that I'm not convinced that this is the definitive way forward, though I'm not ruling it out.

1. I believe the problem here is one of knowledge transfer or documentation. Even if we split the verification, devs would need to know how it works to properly create tasks and update configs for it. The fact that your proposal potentially includes two different worker implementation rollouts for a single cot change means we'd have to make sure devs know about that.

1. I'm only seeing a portion of what you're doing, but I'm seeing a lot of policy and process change proposals from you. I love the enthusiasm. I was once in your position: one of a small (one or two people, in my case) group of relengers focused on mobile, so my use cases and problem sets required different or faster solutions than the desktop majority of the team. Sometimes I was right, and my proposals ended up benefiting the whole team, once we finally rolled them out (sometimes that took years). Sometimes the team was right: because I was focused on a corner of the whole set of projects, I didn't have context for why things were a certain way, and changing things to suit me would be suboptimal, the wrong timing, or even counterproductive for the rest of the team and the project in general. Roles have changed, and I have a larger overall view now. Though I understand where you're coming from, I wonder if you spent comparable amounts of time learning how things work and why as you do proposing changes to our processes and policies, you may have a better perspective of which proposals are good and timely, and which only seem right when you focus on a corner of releng's responsibilities. Similarly, if you state the problem, the specifics of when it comes up, and the urgency of solving that problem, we may be able to help solve the problem in a more collaborative way than needing to head off proposals for change that we see as potentially untimely, unneeded, or even the wrong approach.

escapewindow:
This proposal could be part of our discussions in https://trello.com/c/2ZFwq1of/2-chain-of-trust-future . We should invite the sec team and taskcluster team.

escapewindow:
First off, it looks like my previous comment conveyed the message that you should not write proposals until you have a certain level of experience, and that was not my intention. I apologize for that. Everyone should be able to propose ideas.

Second, I've been giving this more thought. How about:

- We move `cot.verify` into a standalone tool or module
- We add support for running cot verification in all release-related worker implementations. (At some point soon this may just be generic-worker and scriptworker.)
- For release graphs, the worker verifies cot before running any important task. The tasks that have the capability of modifying release artifacts are all important. Any cot verification failure means the worker doesn't run the task.

This means that:

- We get faster feedback about what's broken, e.g. the decision task would fail for an official release graph off an untrusted repo or commit. The build would fail if any of its upstream artifacts were modified at rest. We don't have to wait for a scriptworker task to run to detect it.
- There are fewer places where the chain can be broken.

This is [bug 1492618, end-to-end cot verification](https://bugzilla.mozilla.org/show_bug.cgi?id=1492618), which is a long-standing bug that is still in-progress.

In end-to-end cot verification, we have the option of trusting upstream verification: if the chain hasn't broken until the current task, we can assume the chain is a valid release graph up to this point, and we can just verify the current task's direct inputs. This would save runtime compute cycles.

We also have the option of continuing to verify the upstream tasks in the chain. This would help catch cases where an upstream worker is compromised, as long as that compromise is detectable downstream. For instance, if a decision worker is compromised, it could create a graph and say that the commit is a valid release commit in its cot artifact. However, if the downstream workers verify that the commit is a valid release commit, they would catch the malicious graph. This is not foolproof, but the additional runtime cost may be worth it.

In both of the above cases we'll need the capability of verifying all of the important tasks in the `cot.verify` module or tool.

What do you think? I very much want to see end-to-end cot as a thing, and would very much welcome additional support here.

mitchhentges:
Interesting, I've got three questions, though:

> We add support for running cot verification in all release-related worker implementations. (At some point soon this may just be generic-worker and scriptworker.)

To confirm - `generic-worker` _will_ support decision task verification? Is decision-task verification generic enough, since it depends on digging into `env` variables?

> In end-to-end cot verification, we have the option of trusting upstream verification: if the chain hasn't broken until the current task, we can assume the chain is a valid release graph up to this point, and we can just verify the current task's direct inputs. This would save runtime compute cycles.

As a semantic simplification, is this technique (if chain is correct to current task, assume it's valid so far) equivalent to (one-time decision task reconstruction verification) + (once-per-task artifacts are not tampered verification)?

> We also have the option of continuing to verify the upstream tasks in the chain.

I understand the purpose of doing this (handling the case where an upstream worker is compromised), though I'm not sure it's effective?
If an upstream worker is compromised, then a Chain of Trust signing key is also compromised.
Then, even if we do task-reconstruction downstream, artifacts can be injected and signed with the compromised key. Does re-doing full-chain verification help us, in this case?

escapewindow:
> Interesting, I've got three questions, though:
> 
> > We add support for running cot verification in all release-related worker implementations. (At some point soon this may just be generic-worker and scriptworker.)
> 
> To confirm - `generic-worker` _will_ support decision task verification? Is decision-task verification generic enough, since it depends on digging into `env` variables?

`docker-worker` is going to be EOLed, and all of those tasks moved to `generic-worker`. This has been in progress for the past few quarters, and aiui it'll happen Soon, once `generic-worker` gets docker support and we iron out the bugs. Successful decision tasks (and other docker-worker tasks on generic-worker) are a blocking requirement.

If this doesn't happen, we'll have to add the verification support to docker-worker as well.

> > In end-to-end cot verification, we have the option of trusting upstream verification: if the chain hasn't broken until the current task, we can assume the chain is a valid release graph up to this point, and we can just verify the current task's direct inputs. This would save runtime compute cycles.
> 
> As a semantic simplification, is this technique (if chain is correct to current task, assume it's valid so far) equivalent to (one-time decision task reconstruction verification) + (once-per-task artifacts are not tampered verification)?

We also check all the intervening tasks between the decision task and the scriptworker task, so it's a superset. However, this is not my preferred solution between the two.

> > We also have the option of continuing to verify the upstream tasks in the chain.
> 
> I understand the purpose of doing this (handling the case where an upstream worker is compromised), though I'm not sure it's effective?
> If an upstream worker is compromised, then a Chain of Trust signing key is also compromised.
> Then, even if we do task-reconstruction downstream, artifacts can be injected and signed with the compromised key. Does re-doing full-chain verification help us, in this case?

As I mentioned, it helps us if and when the compromise is not full. If there is anything detectable downstream, we can catch it. If the compromise is done in a way that the downstream can't detect anything, then we won't catch it. In my mind this is still a security improvement, doesn't carry any added code costs (we have to be able to verify every task in the chain and the commit etc anyway, we're just running that verification in more places), and has small runtime costs, so I strongly prefer this.

mitchhentges:
Closing this since we had discussion via Zoom :tada:
TL;DR:
* This approach isn't the most effective, so this RFC is no longer necessary
* There _is_ still potential improvements on failing task groups sooner if the decision task won't pass CoT ([ticket](https://github.com/mozilla-releng/scriptworker/issues/348))
* Chain of Trust is complicated, and maybe potentially it's possible to simplify it without reducing security coverage, but such an approach is unknown atm. Hopefully [end-to-end CoT verification will make this :100:](https://bugzilla.mozilla.org/show_bug.cgi?id=1492618)

PR #29, opened 2019-07-19
Adds "class objects as arguments" RFC

mitchhentges:
This RFC doesn't have a specific implementation goal, Johan and I were discussing architecture on an API boundary and haven't reached a satisfying conclusion, so we opted to tap into the insight of more of the team :smile:

We're requesting reviews from ideally at least Mihai and Aki because Mihai is closest to Releng mobile, and Aki is really familiar with the `scriptworker` ecosystem.
[Rendered.](https://github.com/mitchhentges/releng-rfcs/blob/class-objects-as-arguments/rfcs/0029-class-objects-as-arguments.md)

### DIFF
--- a/rfcs/0029-class-objects-as-arguments.md
+++ b/rfcs/0029-class-objects-as-arguments.md
@@ -1,5 +1,5 @@
-# RFC <number> - Class Objects as Arguments
-* Comments: [#<number>](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/<number>)
+# RFC 0029 - Class Objects as Arguments
+* Comments: [#29](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/29)
 * Proposed by: @mitchhentges
 
 # Summary
--- a/rfcs/0029-class-objects-as-arguments.md
+++ b/rfcs/0029-class-objects-as-arguments.md
@@ -46,7 +46,7 @@ push_apk(                                           |       push_apk(
 **Benefits of using classes to represent intent:**
 * Expected parameters are more explicit and self-documenting - you don't need to depend on docs to tell you that 
 `credentials` is only needed when `contact_google_play=True`
-* Parameters that are related (like `credentials` and `service_account`) are grouped [note 1](#note-1-using-a-dictionary-instead-of-an-object-as-a-parameter)). This is especially
+* Parameters that are related (like `credentials` and `service_account`) are grouped ([there's a note on using a `dict` instead](#note-1-using-a-dictionary-instead-of-an-object-as-a-parameter)). This is especially
 useful when expressing intent, as in "I want to do `<x>`, and here's the associated configuration for doing `<x>`."
 * IDEs can use the class definition to provide Intellisense
 
--- a/rfcs/xxxx-class-objects-as-arguments.md
+++ b/rfcs/xxxx-class-objects-as-arguments.md
@@ -1,0 +1,110 @@
+# RFC <number> - Class Objects as Arguments
+* Comments: [#<number>](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/<number>)
+* Proposed by: @mitchhentges
+
+# Summary
+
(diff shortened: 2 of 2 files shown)


escapewindow:
This seems like a design decision, and I'm unconvinced we need to standardize this across releng. I agree types can be beneficial, yet I also believe that python's flexible typing is an asset. I don't think this is a big enough win to mandate the change unless the major stakeholders all agree it's the best path forward.

This also seems like it doesn't have anywhere near initial consensus, so an rfc is probably the wrong choice.

mitchhentges:
Hey, good points, I'm definitely not looking to standardize this across releng.
However, I needed a platform on which further discussion could occur and where inline comments could be made.
Emails don't allow code formatting, and Slack isn't great for inline comments and responses.

So, I'm using the RFC repo less because this is a releng-wide proposal for standardization, but more because Github provides one of the best discussion platforms, and our RFC repo was the most relevant place to have the discussion

mozbhearsum:
I don't want to pile on too hard, but I also don't think an RFC is the right place for this. Is there a reason this discussion doesn't make sense to stay in the mozapkpublisher repo? If a wider audience is wanted/needed perhaps a link to that could be sent around?

mitchhentges:
@mozbhearsum I agree that this isn't a good fit for an RFC, but where does it make more sense?
My needs for this kind of discussion are:

* Individual discussion points should be possible to annotate inline (Github PRs do a great job of this)
* Once the discussion is resolved, it doesn't litter `mozapkpublisher` with an obsolete file (so, it doesn't make sense to PR this into `mozapkpublisher`)
* It should be possible to format the discussion to make it easier to understand (Github markdown does this well)
* The discussion should be accessible to the wider team so additional reviewers can be brought in as necessary
* The discussion wasn't resolvable with a Zoom meeting, so it needs to be possible to work on asynchronously over a larger time period (days)

Based on these constraints, `releng-rfcs` made the most sense, but it's still an awkward fit. Is there a better workflow for this kind of chat?

mozbhearsum:
If it doesn't fit within the context of an existing PR or Issue in another repo, maybe a Google Doc is a good option, and then it can be linked to from any relevant places?

mitchhentges:
Google docs don't allow as convenient formatting as Github markdown, but that does sound better overall. I'll keep that in mind for the next discussion.
Once this one is complete, I'll close the PR rather than landing

escapewindow:
+1, or a mozapkpublisher issue.

mitchhentges:
The downside of making it a `mozapkpublisher` issue is that we can have inline annotations, you need to do the whole "quote and respond" technique, which makes it hard to have different conversation threads

escapewindow:
Issue linking to a gist?

mitchhentges:
I don't believe that you can leave inline comments on a gist

tomprince:
While this discussion doesn't make sense as something that should land in this repository, it isn't obvious that it shouldn't be used for this kind of discussion (with the end result being the PR is closed), as long as it is clear that that is what is happening.

tomprince:
(Perhaps it would be better as a draft PR, to make landing blocked, as well).

mozbhearsum:
> While this discussion doesn't make sense as something that should land in this repository, it isn't obvious that it shouldn't be used for this kind of discussion (with the end result being the PR is closed), as long as it is clear that that is what is happening.

I agree it isn't obvious or a given, but it's a very notable change to use this repo as a mailing list or discussion board instead of its original purpose. The downside to doing so is that it goes from something with relatively low traffic and high importance, to something high traffic that people will necessarily stop treating with the same importance. If the main problem we're trying to solve is "good UX for arbitrary discussions" I would suggest that we create a repo for that purpose, use Google Docs, or consider using whatever IRC ends up getting replaced with, assuming it addresses the need.

mitchhentges:
Mihai responded [here](https://github.com/mozilla-releng/mozapkpublisher/pull/213#issuecomment-514767189), closing

PR #31, opened 2020-05-10
Automate running ci-admin on push by running it in cloudops' jenkins.

tomprince:
[Rendered](https://github.com/tp-tc/releng-rfcs/blob/ci-admin-automation/rfcs/0031-ci-admin-automation.md)

@mozilla-releng/releng 
@mozilla-releng/relops 
@djmitche @imbstack @petemoore
@moz-hwine @ajvb
@edunham @sciurus

### DIFF
--- a/rfcs/0031-ci-admin-automation.md
+++ b/rfcs/0031-ci-admin-automation.md
@@ -1,0 +1,54 @@
+# RFC 31 - Automate running ci-admin on push. 
+* Comments: [#31](https://github.com/mozilla-releng/releng-rfcs/issues/31)
+* Proposed by: [@tomprince](https://github.com/tomprince)
+
+# Summary
+
+Automate running ci-admin on push by running it in cloudops' jenkins.
+
+## Motivation
+
+- Running ci-admin manually adds extra friction to the process of deploying changes,
+  and can cause confusion when a change is not deployed after landing (either the person 
+  landing not running or not being able to run ci-admin).
+- Running ci-admin manually requires that people have sufficient scopes to apply
+  changes. This means that anybody that can run ci-admin has enough scopes run
+  tasks on level-3 workers and modify level-3 indexes. This makes it harder to
+  reason about the security properties of the system, as we need to take into
+  account people running tasks on those worker directly.
+
+# Details
+
+- The commit access of the [ci-admin](https://hg.mozilla.org/ci/ci-admin) and
+[ci-configuration](https://hg.mozilla.org/ci/ci-configuration) repositories will
+be restricted to a new scm group (`scm_firefox_ci`) and will only be able to be
+pushed to via lando.
+
+  - There will be an escape hatch of putting `MANUAL PUSH` in the commit
+    message, along with a reason, that will trigger secops' audit pipeline.
+
+- The scopes that releng and relops currently have will be restricted to those
+  they need for day-to-day work, and the scopes to run ci-admin will be
+  restricted to the root scope that lives in cloudops' infrastructure.
+
+- A push to ci-admin or ci-configuration will trigger a job in cloudops'
+  jenkins instance that will run ci-admin with the taskcluster root client,
+  after having run ci-admin check.
+
+# Open Questions
+
+- What scopes do releng and relops need for day-to-day work? Since we currently
+  have root-equivalent scopes, we have not needed to figure out what scopes we
+  need otherwise.
+- We need to run `ci-admin` when `.taskcluster.yml` is updated in various repos
+  (to update action hooks). We'll need a way to trigger this, or move away from
+  embedding `.taskcluster.yml` in the hooks.
+
+# Implementation
+
+* [Bug 1619470](https://bugzilla.mozilla.org/show_bug.cgi?id=1619470): Run ci-admin in cloudops's jenkins instance
+* [Bug 1632009](https://bugzilla.mozilla.org/show_bug.cgi?id=1632009): Manage taskcluster clients via ci-admin
+* [Bug 1645168](https://bugzilla.mozilla.org/show_bug.cgi?id=1645168): Restrict pushes to ci-config and ci-admin repos to via lando
+* [Bug 1645170](https://bugzilla.mozilla.org/show_bug.cgi?id=1645170): Restrict pushes to ci-config and ci-admin repos to the scm_firefoxci group.
+* [Bug 1645178](https://bugzilla.mozilla.org/show_bug.cgi?id=1645178): Please monitor pushes to ci/ci-configuration and ci/ci-admin repositories with `MANUAL PUSH` in the commit message.
+


nthomas-mozilla:
I'm in favour of doing this as it will tighten up our processes, and greatly simplify deployment of changes.  It does look a bit like you have need a fixup pass for typos, and a hanging start of sentence in Open Questions, so I encourage you to look at that.

Could you please elaborate on two issues:
* sometimes changes to both ci-admin and ci-configuration are needed, and it's not clear to me how to handle the potentially-broken intermediate state where one has landed but not the other. Can lando land a stack to multiple repos ? Otherwise, some options include mandating backwards compatible changes in ci-admin, or documenting required order if the intermediate will fail to apply
* how this will change the situation for "staging" config ? I mean all the manually applied changes which people are testing on the prod CI. This situation makes has made me a bit uncomfortable because I don't know how it's getting there, and there is the chance of affecting production without meaning to just by having a stale checkout or misclicking in the UI. We might be able to support creating extra pools/workers in a dedicated staging file, and enforce zero overlap with any other configuration with tests. With a further hook exception these changes could land without review and be applied automatically. For: lets us track what's happening; against: adds noise in the commit log.

srfraser:
> I'm in favour of doing this as it will tighten up our processes, and greatly simplify deployment of changes.

+1 This will be a great change, and simplify things enormously. Especially as it means the running configuration should never be local-only, and reverting changes will be visible to others as backouts or new patches.

>     * sometimes changes to both ci-admin and ci-configuration are needed, and it's not clear to me how to handle the potentially-broken intermediate state where one has landed but not the other. Can lando land a stack to multiple repos ? Otherwise, some options include mandating backwards compatible changes in ci-admin, or documenting required order if the intermediate will fail to apply

Being backwards compatible seems like a good goal, although we should also work out how long it takes Jenkins to deploy a new worker so we know how to avoid the race (assuming the task setup for ci-configuration isn't going to include 'hg clone ci-admin'). It's tempting to try and add dependencies or ci-admin version checks, but I worry they'll be unnecessary complication and not add much compared to human planning for those situations.

djmitche:
To Nick's second point, one option might be to have a way to define resources that are ignored in staging by some kind of regex on their id.  (Ignored meaning, ci-admin won't delete, update, or create them).  You could prime that list with patterns like `/bug[0-9]{7}$/` and then add others as necessary (e.g., `^WorkerType=.*-alibaba$` when we start the Alibaba migration).

djmitche:
The advantage to the `/bug[0-9]{7}/` strategy is that it's easy to answer the questions "what is this for?" (see bug) and "is this still needed?" (is the bug still open) without having to ask anyone.

djmitche:
Also "Auto*t*mate" :)

rfcs/0031-ci-admin-automation.md
```diff
@@ -0,0 +1,47 @@
...
+
+# Details
+
+- The commit access of the [ci-admin](https://hg.mozilla.org/ci/ci-admin) and
+[ci-configuration](https://hg.mozilla.org/ci/ci-configuration) repositories will
+be restricted to a new scm group (`scm_firefox_ci`) and will only be able to be
+pushed to via lando.
+
+  - There will be an escpae hatch of putting `MANAUAL PUSH` in the commit
+    message, along with a reason, that will trigger secops' audit pipeline.
```
ajvb, line 28: I like this idea and having an alert from the fraud pipeline for this is definitely doable, but I'd like to understand this a bit more. How will the "MANUAL PUSH" process differ from what currently exists? Will it still require some sort of review?

escapewindow: I'm guessing `MANUAL PUSH` pushes are likely pushed without Phabricator review, in case one of us needs to push an emergency fix when no one is around to review.

mozbhearsum: I think it's worth asking the question of whether or not this should be allowed at all. We've got enforced, multiple, signoffs in a couple of places already, and an escalation chain. Or perhaps we should go with this, and remove it if it's clear it's rarely if ever used?

escapewindow: I can think of plenty of times when there's some tree-closing bustage and I'm the only Relenger online, so no reviews are available. I would argue that having the source control change visible and auditable is still a vast security improvement over today's apply-arbitrary-changes-at-will situation. If we don't have the `MANUAL PUSH` solution, we still need some way to apply changes when everyone else is offline. Paging `{cloud|sec}ops` may be a solution, though they may not have the context to review a change properly, so this may be similar to a `MANUAL PUSH` that alerts secops. We could also start rotating pagerduty on our team, in case we want off-hours reviews. Of the above solutions, I like the `MANUAL PUSH` solution best, but there may be other opinions or other solutions I haven't listed.

tomprince: https://bugzilla.mozilla.org/show_bug.cgi?id=1515119 has some context (which is where this was implemented for mozilla-central and related branches).

I'll note that review isn't required by technical means on mozilla-central or these repositories (either by lando or pushing manually), though lando does require you to acknowledge that things haven't been reviewed before landing. I'm not proposing to change that. 

The way it works for m-c is that you can only push to those branches via lando, (or if you have the `scm_direct_push` group); but you can bypass that check by adding `MANUAL PUSH: <reason>` in the commit message, which gets logged to sentry (and I'm proposing triggers audit pipeline for ci repositories).

As @escapewindow indicated, we've had to deal with issues out-of-hours and having a way to deal with that is useful, and this provides better tracking and alerting for that than is currently in place. We can certainly explore adding more restrictions in the future, but I'd like to keep that out of scope of this proposal.

tomprince:
>     * sometimes changes to both ci-admin and ci-configuration are needed, and it's not clear to me how to handle the potentially-broken intermediate state where one has landed but not the other. Can lando land a stack to multiple repos ? Otherwise, some options include mandating backwards compatible changes in ci-admin, or documenting required order if the intermediate will fail to apply

Lando doesn't currently support landing things across repos. I think the main case to worry about is if landing things out of sync generate a *bad* configuration, rather than failing to apply, but I suspect that isn't the common case. I think what would typically happen is the change to one is landed, which triggers an deploy that fails, followed by landing the other, which then successfully applies (with the possibility to race jenkins to skip the failed application).

There are probably better ways to handle this, but I think that is something we can iterate on over time.

>     * how this will change the situation for "staging" config ? I mean all the manually applied changes which people are testing on the prod CI.

My inclination is to require landing things to get them deployed. Another option would be to carve out some space that ci-admin doesn't touch, and allow people to make changes to that directly, but I suspect figuring out the right security restrictions would be somewhat tricky.

The other thing we can do is improve the ability to test changes on the staging cluster before landing. I'm uncertain if we want to run ci-admin automatically there, but running manually will be an option there in any case.

tomprince:
> Have we considered how to force a ci-admin run on `.taskcluster.yml` change, to rebuild hooks? Or are we going to deprecate embedding `.taskcluster.yml` in the hooks sooner?

I'll add this to the open questions. I suspect we can get cloudops to trigger a run for us (or this may be something those of us with access to jenkins can run).

escapewindow:
Lando doesn't currently support landing things across repos. I think the main case to worry about is if landing things out of sync generate a bad configuration, rather than failing to apply, but I suspect that isn't the common case. I think what would typically happen is the change to one is landed, which triggers an deploy that fails, followed by landing the other, which then successfully applies (with the possibility to race jenkins to skip the failed application).

Spitballing: what if we had a min-admin-revision in ci-configuration? ci-admin could make sure its history contains that revision as a parent revision, or it'll bail.

Also, I just accidentally edited your comment (I meant to quote-reply). Wondering if we should disallow editing each other's comments, or if I should just be more careful =\

rfcs/0031-ci-admin-automation.md
ajvb: I do agree with @escapewindow that even without multiple signoff or paging another mozilla employee this will be an improvement, but we should make sure we have a plan in place for further tightening the security controls of the `MANUAL PUSH` process. May be worth having a short meeting to discuss various options? I think I need to understand more about what exactly ci-admin controls before I can confidently suggest one option over another.

davehouse: +1 this will be better than now, and I'd like to hear about the ci-admin impact to prepare for if I'm ever in the position of needing to r+ a `MANUAL PUSH.` I can read the files, but I may not understand the consequences of _any_ change.

mozbhearsum: I don't see multiple sigonff as ensuring that a thing is this fully and properly reviewed. It's more about making sure that one person's ldap account can't make infrastructure changes alone. For that use case, a sanity check is all that's needed.

I also agree that `MANUAL PUSH` is better than the current state, so improving on that doesn't need to block this.

rfcs/0031-ci-admin-automation.md
```diff
@@ -0,0 +1,52 @@
+# RFC 31 - Automate running ci-admin on push. 
+* Comments: [#31](https://github.com/mozilla-releng/releng-rfcs/issues/31)
+* Proposed by: [@tomprince](https://github.com/tomprince
```
hwine, line 3: missing close paren ;)

rfcs/0031-ci-admin-automation.md
```diff
@@ -0,0 +1,52 @@
...
+- We need to run `ci-admin` when `.taskcluster.yml` is updated in various repos
+  (to update action hooks). We'll need a way to trigger this, or move away from
+  embedding `.taskcluster.yml` in the hooks.
+
+# Implementation
+
+* [Bug 1619470](https://bugzilla.mozilla.org/show_bug.cgi?id=1619470): Run ci-admin in cloudops's jenkins instance
+* [Bug 1632009](https://bugzilla.mozilla.org/show_bug.cgi?id=1632009): Manage taskcluster clients via ci-admin
+* Bug XXXXXXX: Restrict pushes to ci-admin and ci-configuration
+
```
hwine, line 52: Please file bug for adding fraud pipeline check [here](https://bugzilla.mozilla.org/enter_bug.cgi?assigned_to=nobody%40mozilla.org&bug_ignored=0&bug_severity=--&bug_status=NEW&bug_type=task&cf_fx_iteration=---&cf_fx_points=---&cf_status_firefox76=---&cf_status_firefox77=---&cf_status_firefox78=---&cf_status_firefox79=---&cf_status_firefox_esr68=---&cf_tracking_firefox76=---&cf_tracking_firefox77=---&cf_tracking_firefox78=---&cf_tracking_firefox79=---&cf_tracking_firefox_esr68=---&cf_tracking_firefox_relnote=---&cf_tracking_firefox_sumo=---&component=Fraud%20Detection&contenttypemethod=list&contenttypeselection=text%2Fplain&defined_groups=1&filed_via=standard_form&flag_type-37=X&flag_type-607=X&flag_type-708=X&flag_type-721=X&flag_type-737=X&flag_type-748=X&flag_type-787=X&flag_type-800=X&flag_type-803=X&flag_type-846=X&flag_type-864=X&flag_type-936=X&flag_type-941=X&form_name=enter_bug&maketemplate=Remember%20values%20as%20bookmarkable%20template&op_sys=Unspecified&priority=--&product=Cloud%20Services&rep_platform=Unspecified&target_milestone=---&version=unspecified), and add to this list.

Also file the bug for restriction pushes, please.

tomprince:
The RFC is in final-comment mode until Friday, meaning that we believe
there is agreement and would like to hear any objections before deciding to
proceed with implementation.  Please speak up on the issue if you believe
we should not proceed as proposed.

rfcs/0031-ci-admin-automation.md
```diff
@@ -0,0 +1,52 @@
+# RFC 31 - Automate running ci-admin on push. 
+* Comments: [#31](https://github.com/mozilla-releng/releng-rfcs/issues/31)
+* Proposed by: [@tomprince](https://github.com/tomprince
+
+# Summary
+
+Automate running ci-admin on push by running it in cloudops' jenkings.
```
escapewindow, line 7: * jenkins

rfcs/0031-ci-admin-automation.md
```diff
@@ -0,0 +1,52 @@
...
+  account people running tasks on those worker directly.
+
+# Details
+
+- The commit access of the [ci-admin](https://hg.mozilla.org/ci/ci-admin) and
+[ci-configuration](https://hg.mozilla.org/ci/ci-configuration) repositories will
+be restricted to a new scm group (`scm_firefox_ci`) and will only be able to be
+pushed to via lando.
+
+  - There will be an escape hatch of putting `MANAUAL PUSH` in the commit
```
escapewindow, line 27: * MANUAL

PR #33, opened 2021-04-13
Release for Mozilla: Shared Trust Domain & Workers

bhearsum:
[rendered](https://github.com/bhearsum/releng-rfcs/blob/generic-worker-pools/rfcs/0033-shared-trust-domain.md)

### DIFF
--- a/rfcs/0033-shared-trust-domain.md
+++ b/rfcs/0033-shared-trust-domain.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-Build and maintain a shared Trust Domain, Workers, and Scriptworkers on the Firefox CI cluster that any Mozilla project can use. (Browser products will remain in their existing - separate - trust domain.)
+Build and maintain a shared Trust Domain, Workers, and Scriptworkers on the Firefox CI cluster that any Mozilla project can use. Browser products will remain in their existing - separate - trust domain.
 
 ## Motivation
 
@@ -15,29 +15,59 @@ One of the barriers to entry for using Taskcluster is waiting on RelEng to creat
 We will create a new trust domain and workers that are generally available for Mozilla employees and trusted volunteers to use. Specifically:
 
 * A new Trust Domain (`mozilla`) that is not tied to a specific project or product
-* New Workers for builds on Linux, macOS 11.0, and Windows Server 2012
+* New Workers for builds on Linux, macOS 10.15, and Windows Server 2012
+    * A RelEng maintained Docker image will be provided for Linux
     * These will be created under a new `mozilla-1` provisioner
-* New Workers for tests on Linux (through developer provided Docker images), macOS 11.0, and Windows 10
+* New Workers for tests on Linux, macOS 10.15, and Windows 10
+    * A RelEng maintained Docker image will be provided for Linux
     * These will be created under a new `mozilla-t` provisioner
-* New Scriptworker instances forsigning and mac-signing
+* New Scriptworker instances for signing and mac-signing
 	* These will be created under the existing `scriptworker-k8s` and `scriptworker-prov-v1` provisioners
 	* Workers will be prefixed with `mozilla-t-`
+    * mac-signing will run 10.14, like our other mac-signing workers (there's no known reason to upgrade)
 
-Notably, we are only concerned with level 1 workers at this time, which means we can ignore things like scriptworkers that are only used when shipping. Level 3 workers will be dealt with at a later stage.
+Notably, we are only concerned with level 1 workers at this time, which means we can ignore things like scriptworkers that are only used when shipping. Level 3 workers will be dealt with at a later stage, and most likely will not use a shared trust domain or workers across projects.
 
-Access to create and manage tasks on these new workers will be granted to anyone with `scm_level_1`.
+Access to create and manage tasks on these new workers will be granted to anyone with `scm_level_1_github` or `scm_level_1`.
 
 Going forward, we will ensure workers for other supported build or target platforms are added to this pool. (For example, when we add support for scheduling iOS tests in Taskcluster, that will be made available in the `mozilla-t` provisioner as well.)
 
-# Open Questions
+## List of Pools
 
-* Are we happy with the new trust domain name & provisioners for the workers?
-* Where did we get the macOS hardware for the build, test, and signing pools?
-	* New or pull from existing pools?
-	* How many machines do we need in each hardware pool?
-* Is macOS 11.0 the right version to use for build and test?
-* Are there other test platforms or scriptworkers we should support?
-* Is `scm_level_1` the right group to use, or do we need a new one for this purpose?
+Pool ID                   | Purpose
+====================================================================================
+mozilla-1/linux           | Linux jobs
+mozilla-1/linux-highcpu   | Linux jobs requiring more CPU resources
+mozilla-1/win2012         | Windows Server 2012 jobs
+mozilla-1/win2012-highcpu | Windows Server 2012 jobs requiring more CPU resources
+mozilla-1/win10           | Windows 10 jobs
+mozilla-1/macos-bigsur    | macOS 10.15 jobs
+mozilla-t/signing         | Non-mac signing jobs
+mozilla-t/mac-signing     | Mac signing jobs
+
+## Hardware Machine Allocation
+
+We will need hardware for 3 different pools, which will be allocated as noted below:
+* 2 machines for macOS signing, running allocated from the existing production Firefox pool
+* 3 machines for macOS builds, allocated from TBD
+* 3 machines for macOS tests, allocated from TBD
+
+When additional workers are needed in the future, they will be allocated from TBD.
+
+## `v3` Taskcluster index format
+
+The current `v2` index format only includes repository names as an identifier (not user or organization). This is generally not an issue for any of current trust domains (because they generally only support one project), but in this new pool where we have N projects, it introduces the potential for collisions or pollution between them. To ensure this isn't an issue we will introduce a new `v3` index format that includes the repository location in its path as well - including both domain and path. Examples include:
+* `index.mozilla.v3.github.com.mozilla-mobile.mozilla-vpn-client.branch.main.latest.taskgraph.decision`
+* `index.gecko.v3.hg.mozilla.org.releases.mozilla-beta.latest.firefox.decision`
+
+This will require changes to a few things to support the new format:
+* build-decision
+* scriptworker
+* taskgraph
+
+Existing users of Taskcluster will not be required to upgrade to the v3 format.
+
+# Open Questions
 
 # Implementation
 
--- a/rfcs/0033-shared-trust-domain.md
+++ b/rfcs/0033-shared-trust-domain.md
@@ -21,7 +21,7 @@ We will create a new trust domain and workers that are generally available for M
     * These will be created under a new `mozilla-t` provisioner
 * New Scriptworker instances forsigning and mac-signing
 	* These will be created under the existing `scriptworker-k8s` and `scriptworker-prov-v1` provisioners
-	* Workers will be prefixed with `mozilla-1-`
+	* Workers will be prefixed with `mozilla-t-`
 
 Notably, we are only concerned with level 1 workers at this time, which means we can ignore things like scriptworkers that are only used when shipping. Level 3 workers will be dealt with at a later stage.
 
--- a/rfcs/0033-shared-trust-domain.md
+++ b/rfcs/0033-shared-trust-domain.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-Build and maintain a shared Trust Domain, Workers, and Scriptworkers on the Firefox CI cluster that any Mozilla project can use.
+Build and maintain a shared Trust Domain, Workers, and Scriptworkers on the Firefox CI cluster that any Mozilla project can use. (Browser products will remain in their existing - separate - trust domain.)
 
 ## Motivation
 
@@ -46,4 +46,3 @@ implementation through to completion>
 
 * <link to tracker bug, issue, etc.>
 * <...>
-
--- a/rfcs/0033-shared-trust-domain.md
+++ b/rfcs/0033-shared-trust-domain.md
@@ -19,7 +19,7 @@ We will create a new trust domain and workers that are generally available for M
     * These will be created under a new `mozilla-1` provisioner
 * New Workers for tests on Linux (through developer provided Docker images), macOS 11.0, and Windows 10
     * These will be created under a new `mozilla-t` provisioner
-* New Scriptworkers instances forsigning and mac-signing
+* New Scriptworker instances forsigning and mac-signing
 	* These will be created under the existing `scriptworker-k8s` and `scriptworker-prov-v1` provisioners
 	* Workers will be prefixed with `mozilla-1-`
 
--- a/rfcs/0033-shared-trust-domain.md
+++ b/rfcs/0033-shared-trust-domain.md
@@ -1,0 +1,49 @@
+# RFC 33 - Release for Mozilla: Shared Trust Domain & Workers
+* Comments: [#33](https://github.com/mozilla-releng/releng-rfcs/pull/33)
+* Proposed by: @bhearsum
+
+# Summary
+
+Build and maintain a shared Trust Domain, Workers, and Scriptworkers on the Firefox CI cluster that any Mozilla project can use.
+
+## Motivation
+
+One of the barriers to entry for using Taskcluster is waiting on RelEng to create and deploy a new Trust Domain and Worker for a new project. Even when this takes less than a day to do (and it often takes longer), it's still something that needs to be waited on, and is slower than using CircleCI or Github Actions.
+
+# Details
+
+We will create a new trust domain and workers that are generally available for Mozilla employees and trusted volunteers to use. Specifically:
+
+* A new Trust Domain (`mozilla`) that is not tied to a specific project or product
+* New Workers for builds on Linux, macOS 11.0, and Windows Server 2012
+    * These will be created under a new `mozilla-1` provisioner
+* New Workers for tests on Linux (through developer provided Docker images), macOS 11.0, and Windows 10
+    * These will be created under a new `mozilla-t` provisioner
+* New Scriptworkers instances forsigning and mac-signing
+	* These will be created under the existing `scriptworker-k8s` and `scriptworker-prov-v1` provisioners
+	* Workers will be prefixed with `mozilla-1-`
+
+Notably, we are only concerned with level 1 workers at this time, which means we can ignore things like scriptworkers that are only used when shipping. Level 3 workers will be dealt with at a later stage.
+
+Access to create and manage tasks on these new workers will be granted to anyone with `scm_level_1`.
+
+Going forward, we will ensure workers for other supported build or target platforms are added to this pool. (For example, when we add support for scheduling iOS tests in Taskcluster, that will be made available in the `mozilla-t` provisioner as well.)
+
+# Open Questions
+
+* Are we happy with the new trust domain name & provisioners for the workers?
+* Where did we get the macOS hardware for the build, test, and signing pools?
+	* New or pull from existing pools?
+	* How many machines do we need in each hardware pool?
+* Is macOS 11.0 the right version to use for build and test?
+* Are there other test platforms or scriptworkers we should support?
+* Is `scm_level_1` the right group to use, or do we need a new one for this purpose?
+
+# Implementation
+
+<once the RFC is decided, these links will provide readers a way to track the
+implementation through to completion>
+
+* <link to tracker bug, issue, etc.>
+* <...>
+


rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Motivation
+
+One of the barriers to entry for using Taskcluster is waiting on RelEng to create and deploy a new Trust Domain and Worker for a new project. Even when this takes less than a day to do (and it often takes longer), it's still something that needs to be waited on, and is slower than using CircleCI or Github Actions.
+
+# Details
+
+We will create a new trust domain and workers that are generally available for Mozilla employees and trusted volunteers to use. Specifically:
+
+* A new Trust Domain (`mozilla`) that is not tied to a specific project or product
```
escapewindow, line 17: Security speaking, we will not isolate caches between projects using the same trust domain, and admins of each project may be able to admin other projects in the same trust domain.

I wonder if we'll soon hit issues with index collisions as well. Since we don't include the organization in project names, if there's a `mozilla-releng/persona` and a `mozilla-mobile/persona` both using the same `mozilla` trust domain, they will be writing to the same `mozilla.v2.persona.*` index namespace. We might not worry about indexes with hashes in them, since hash collisions will be very rare, but indexes with `latest` or datestrings may end up with more noise than signal. (A malicious user with admin in one project could be able to overwrite other projects' hash indexes as well, since the decision task will have scopes to create said indexes and the logic of how to generate the hash is modifiable in-tree.)

We may be able to live with some of these issues at level 1, but we likely want to resolve them before we enable level 3.

I wonder if this will accelerate our adoption of [`projectId`](https://github.com/taskcluster/taskcluster-rfcs/blob/main/rfcs/0163-project-id.md).

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,49 @@
...
+* New Workers for builds on Linux, macOS 11.0, and Windows Server 2012
+    * These will be created under a new `mozilla-1` provisioner
+* New Workers for tests on Linux (through developer provided Docker images), macOS 11.0, and Windows 10
+    * These will be created under a new `mozilla-t` provisioner
+* New Scriptworkers instances forsigning and mac-signing
```
escapewindow, line 22: Nit: sp, `Scriptworker`, `for signing`

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,49 @@
...
+	* These will be created under the existing `scriptworker-k8s` and `scriptworker-prov-v1` provisioners
+	* Workers will be prefixed with `mozilla-1-`
```
escapewindow, line 24: For signing, our pattern has been `{trust_domain}-t-signing` for dep. This is because `{trust_domain}-1` assumes we're running against a level 1 tree. `{trust_domain}-t` runs against levels 1-3. I think both names have their own issues, but convention currently says `-t` for dep signing.

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,49 @@
...
+Access to create and manage tasks on these new workers will be granted to anyone with `scm_level_1`.
+
+Going forward, we will ensure workers for other supported build or target platforms are added to this pool. (For example, when we add support for scheduling iOS tests in Taskcluster, that will be made available in the `mozilla-t` provisioner as well.)
+
+# Open Questions
+
+* Are we happy with the new trust domain name & provisioners for the workers?
+* Where did we get the macOS hardware for the build, test, and signing pools?
+	* New or pull from existing pools?
+	* How many machines do we need in each hardware pool?
```
escapewindow, line 37: We can probably pull several prod mac signers out of the Firefox pool and add them here for dep mac signing.

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,49 @@
...
+* Is macOS 11.0 the right version to use for build and test?
+* Are there other test platforms or scriptworkers we should support?
```
escapewindow, line 39: Githubscript would provide Github tagging/release support. Someday it may support creating PRs or commenting on revisions with task statuses.

Treescript would provide version bumping and pushing to a branch. We may add PR creation here or Githubscript.

Both of the above scripts would require project admins inviting the bot user to their repo with more than read privs. It's quite possible we don't need either for the initial rollout.

Beetmoverscript could be useful, but possibly only when we start doing nightlies or releases.

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,49 @@
...
+* Is `scm_level_1` the right group to use, or do we need a new one for this purpose?
```
escapewindow, line 40: I think it works. I think the main downside is if we kick people out if they haven't pushed to hg.m.o Try in a number of months: I could see needing a similar flag for developers who only work on Github.

rfcs/0033-shared-trust-domain.md
escapewindow: (I can totally see this issue leading to a `v3` index format.)

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,49 @@
+# RFC 33 - Release for Mozilla: Shared Trust Domain & Workers
+* Comments: [#33](https://github.com/mozilla-releng/releng-rfcs/pull/33)
+* Proposed by: @bhearsum
+
+# Summary
+
+Build and maintain a shared Trust Domain, Workers, and Scriptworkers on the Firefox CI cluster that any Mozilla project can use.
```
hwine, line 7:
```suggestion
Build and maintain a shared Trust Domain, Workers, and Scriptworkers on the Firefox CI cluster that any Mozilla project can use. (Browser products will remain in their existing - separate - trust domain.)
```

I think that answering a question many will have up front is a good thing.  Any further clarification would be helpful as well. (e.g. "any _new_ Mozilla project" if that's the case.)

hwine: Related, if there is a planned transition point out of the shared trust domain, that'd be nice to know. E.g. when projects start shipping, will their L3 builders be in this trust domain, or a separate one? Or is that included in "we'll deal with L3 later" comment below?

rfcs/0033-shared-trust-domain.md
hwine: another option here would be to include a UID for the repo. GitHub provides one, which should rule out someone "hijacking" the project by changing the config to pull from `attacker/persona`. But yeah, numerous security issues here.

re: "admins of each project may be able to admin other projects" -- an admin-list-per-project would be preferable (but I'm unaware of how much coding that would take).

rfcs/0033-shared-trust-domain.md
escapewindow: +1, when projects need releases or clearly have sensitive secrets that we don't want accessible by other projects, that's a good candidate for its own trust domain.

rfcs/0033-shared-trust-domain.md
bhearsum: For the purposes of this RFC, I'd like to keep this scoped at level 1. IMO, the benefits outweigh the risks at level 1 (modulo the implementation issues that @escapewindow raised). Level 3 is a different conversation, of course, and I wouldn't be shocked if we end up wanting separate trust domains for that.

rfcs/0033-shared-trust-domain.md
bhearsum: Okay! I thought I had this straight already, but clearly not.

rfcs/0033-shared-trust-domain.md
bhearsum: Do you think we need these for level 1? AFAICT we don't have them for Firefox or Fenix level 1.

rfcs/0033-shared-trust-domain.md
bhearsum: I hadn't thought about the idea of fully moving to a new trust domain at some point - I was thinking more than level 1 would be shared, and level 3 would be per-project (this has some good benefits, especially when we're talking about pools of hardware machines). I could see both working, though.

bhearsum: In any case, this suggestion/clarification is a good one.

escapewindow: Hm, I hadn't thought about using a trusted level 3 trust domain, and a shared level 1 trust domain. I don't think we've tried that. Having said that, modulo implementation details, this may be possible.

rfcs/0033-shared-trust-domain.md
escapewindow: Yeah, if we make it explicit that this shared trust domain is only for level 1, then we *may* want githubscript or treescript in the future to automate creating PRs or commenting on revisions, but we can probably leave them out til we actually have a real world need.

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,48 @@
...
+* A new Trust Domain (`mozilla`) that is not tied to a specific project or product
+* New Workers for builds on Linux, macOS 11.0, and Windows Server 2012
+    * These will be created under a new `mozilla-1` provisioner
+* New Workers for tests on Linux (through developer provided Docker images), macOS 11.0, and Windows 10
+    * These will be created under a new `mozilla-t` provisioner
+* New Scriptworkers instances forsigning and mac-signing
+	* These will be created under the existing `scriptworker-k8s` and `scriptworker-prov-v1` provisioners
+	* Workers will be prefixed with `mozilla-1-`
+
+Notably, we are only concerned with level 1 workers at this time, which means we can ignore things like scriptworkers that are only used when shipping. Level 3 workers will be dealt with at a later stage.
```
escapewindow, line 26: I think this line makes it sound like we'll add level 3 support to this trust domain at some point. Our conversations seem like we're restricting this RFC to level 1 only. Shall we change this line to say something like "We only plan to support level 1 in this trust domain"?

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,48 @@
...
+
+Access to create and manage tasks on these new workers will be granted to anyone with `scm_level_1`.
+
+Going forward, we will ensure workers for other supported build or target platforms are added to this pool. (For example, when we add support for scheduling iOS tests in Taskcluster, that will be made available in the `mozilla-t` provisioner as well.)
+
+# Open Questions
+
+* Are we happy with the new trust domain name & provisioners for the workers?
```
MihaiTabara, line 34: Having a single trust-domain means we may also hit cost/performance conflicts since different projects have different needs. E.g. an application-service type project needs beefy machines to compile Rust, while a light-mobile one will require inexpensive hardware. But I think for level-1 workes, we'll simply have to trust the developers that they'll not abuse the `mozilla-1/b-linux-large` in lieu of`mozilla-1/b-linux` for example. We may want to find a way to separate these better for L3 counterparts.

rfcs/0033-shared-trust-domain.md
bhearsum: Good point; we know those aren't very busy.

rfcs/0033-shared-trust-domain.md
bhearsum: I still think we should investigate the idea of a shared pool for level 3 for smaller/less critical projects, but I'm OK writing this on the assumption that that won't be possible.

rfcs/0033-shared-trust-domain.md
bhearsum: This gets at something I hadn't thought of much actually -- which instances we'd actually support in the pool. My instinct is to say that we should provide at least a memory-optimized, compute-optimized, and (maybe?) storage optimized to give people a choice of what's best for them.

I don't think we should think too much about abuse here. If people are behaving badly that's likely something that needs to be fixed socially rather than technologically. It could be an argument for enhancing our cost reporting, so we can "bill" the projects, or at the very least, correctly be able to report which project is blowing up costs if questions come along.

MihaiTabara: Yeah, agreed this can be a good argument for cost-reporting and ways to "bill" per project. But until then, `memory-optimized`, `compute-optimized` and `storage-optimized` seems like a good start to at least cover a wider range of projects.

rfcs/0033-shared-trust-domain.md
bhearsum: For posterity & the benefit of others, these are two concrete reasons we may want githubscript: https://github.com/mozilla-releng/scriptworker-scripts/issues/319, https://github.com/mozilla-releng/scriptworker-scripts/issues/320

bhearsum: For our purposes here, I don't think we need githubscript right away, as it doesn't support any features that are useful on level 1 IMO. I do think we'll want them at a later date.

rfcs/0033-shared-trust-domain.md
bhearsum: I realize I haven't address @escapewindow's original comment here! It sounds like there's two simple ways to deal with it:
1) Include the Github repo UID somewhere in the index path.
2) Include the Github organization or user somewhere in the index path

IMO 1) is a bit less nice because it makes the indexes less human readable (even if we keep the repo name, an ugly, useless-to-humans, UID will be in there).

@escapewindow - do you think 2) will address your concern?

escapewindow: Yeah, I think the UID may require an api lookup for (1). Maybe, possibly, not sold completely on (1) though it does at least partially address the "I've deleted or renamed my repo and someone else took my old repo name" scenario.

I do think including the Github org or user would work, but that change would:
- potentially break some current assumptions, like where we want forks to be able to populate indexes in the same namespace, just with a `level-1` in the index (though *this* is a bad assumption since forks don't have to have the same repo name!)
- will require changes in the indexes which means build-decision, shipit, CoT verification, various other automation will need to be updated to use the new indexes. This is why I think we want a `v3` index format.

bhearsum: > we want forks to be able to populate indexes in the same namespace, just with a level-1 in the index (though this is a bad assumption since forks don't have to have the same repo name!)

If this is a requirement, it seems like a dealbreaker for my option 2). A `v3` index format seems like no big deal though (just some amount of straightforward change).

hwine: fwiw, you can "easily" find the parent repo of a fork via the api.

escapewindow: Yes, but we don't do so in granting scopes or following indexes.

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,48 @@
...
+One of the barriers to entry for using Taskcluster is waiting on RelEng to create and deploy a new Trust Domain and Worker for a new project. Even when this takes less than a day to do (and it often takes longer), it's still something that needs to be waited on, and is slower than using CircleCI or Github Actions.
+
+# Details
+
+We will create a new trust domain and workers that are generally available for Mozilla employees and trusted volunteers to use. Specifically:
+
+* A new Trust Domain (`mozilla`) that is not tied to a specific project or product
+* New Workers for builds on Linux, macOS 11.0, and Windows Server 2012
+    * These will be created under a new `mozilla-1` provisioner
+* New Workers for tests on Linux (through developer provided Docker images), macOS 11.0, and Windows 10
```
bhearsum, line 20: I was thinking about these separated build & test pools a bit more, specifically in the context of the lack of separation of these things proposed in https://github.com/mozilla-releng/releng-rfcs/pull/36. Specifically, I'm wondering if we should do away with that distinction here and just share general purpose pools. Something like:
```
Pool ID                   | Purpose
====================================================================================
mozilla-1/linux           | Linux jobs
mozilla-1/linux-highcpu   | Linux jobs requiring more CPU resources
mozilla-1/win2012         | Windows Server 2012 jobs
mozilla-1/win2012-highcpu | Windows Server 2012 jobs requiring more CPU resources
mozilla-1/win10           | Windows 10 jobs
mozilla-1/macos-bigsur    | macOS Big Sur jobs
```

I can see why we keep them separate for Gecko (it lets us share test machines for level 1 & level 3 jobs), but I can't think of a reason why we need to do it here. @escapewindow - do you have any thoughts on this?

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,48 @@
...
+
+Going forward, we will ensure workers for other supported build or target platforms are added to this pool. (For example, when we add support for scheduling iOS tests in Taskcluster, that will be made available in the `mozilla-t` provisioner as well.)
+
+# Open Questions
+
+* Are we happy with the new trust domain name & provisioners for the workers?
+* Where did we get the macOS hardware for the build, test, and signing pools?
+	* New or pull from existing pools?
+	* How many machines do we need in each hardware pool?
+* Is macOS 11.0 the right version to use for build and test?
```
bhearsum, line 38: @jmaher - do you have any thoughts on macOS version to use here?

bhearsum: Thanks Joel! I'm not particularly concerned about CPU arch (I don't think there's anything that would use this new pool anytime soon that would be sensitive to it) - mostly I just want to pick the most convenient macOS version from a management standpoint. It sounds like it's pretty clear that that is 10.15.

rfcs/0033-shared-trust-domain.md
bhearsum: I'm going to push a change proposes a `v3` index format. I _think_ that will address this thread, but please correct re-open this thread if I'm wrong.

rfcs/0033-shared-trust-domain.md
bhearsum: Thinking about this a bit more, it's pretty obviously going to be an issue quickly. We have lots of stuff on Github now, and many developers who never touch hg. Unless we can have the autoremoval process take into account Github activity as well, I don't think this is going to be a viable option. I'll talk to IT about that, but for now I'm going to propose a new group on the assumption that it won't be possible.

rfcs/0033-shared-trust-domain.md
bhearsum: I'm descoping this a little bit to only deal with general purpose and high cpu (these are the only types we use for Firefox & Fenix AFAICT).

rfcs/0033-shared-trust-domain.md
bhearsum: I'm going to push a change with the proposed pools; this comment may disappear when I do so, but we can re-open or talk in a new thread.

hwine: > I can see why we keep them separate for Gecko (it lets us share test machines for level 1 & level 3 jobs), but I can't think of a reason why we need to do it here. @escapewindow - do you have any thoughts on this?

Another reason for the separate pools is so we can trust that builders haven't been corrupted by a rogue test run. At least in the case of "release builds" (whatever that means in this context), it would be good to have a high level of trust in the builder.

I know there are other ways to achieve that goal than a separate pool, but if there's a spot we can capture that requirement, that would be good.

bhearsum: Can "rogue builds" corrupt in the same way that "rogue tests" can? Or is it about minimizing the things that can influencing builds?

Either way, separating these out is not terribly difficult, so I don't strong feelings here.

hwine: good catch! we do care about rogue builds in the "accept outside PRs" world. If there is a concept of a "release build", it should include purging the cache of non-release-build artifacts.

escapewindow: Hm, keeping them compute pools rather than build/test pools may make sense for now, as long as we're limited to level 1 (in level 3, we need to guard against tests poisoning release build caches and workers). If we need separation in the future, we can add b- and t- prefixes as needed.

bhearsum: So it sounds like we're all OK with shared pools for level-1, but (almost?) certainly not for level-3 -- which are not part of this proposal.

bhearsum:
I _think_ we're getting close to consensus here. I think the two things left are:
1) Making sure the `v3` index format is sensible, fleshed out enough, and solves the problem it's meant to. @hwine, @escapewindow - what do you think here?
2) Making sure the new LDAP group is doable and accepable. I'm pretty sure it's fine on our side (please speak up if you disagree), and I'm going to have a chat with IT about their side.

If there's anything else I'm missing, please let me know.

hwine:
> I _think_ we're getting close to consensus here. I think the two things left are:
> 
>     1. Making sure the `v3` index format is sensible, fleshed out enough, and solves the problem it's meant to. @hwine, @escapewindow - what do you think here?

@ajvb -- I'll defer to you on the `v3` index - that's deeper TaskCluster stuff than I'm comfortable with.

ajvb:
@bhearsum In regards to the `v3` index format, are there specific security concerns around that format that you have?

My (little) familiarity with indexes in Taskcluster leads me to believe there isn't much of security concern unless we are taking about one project polluting another project, which isn't really related to the index format.

bhearsum:
> My (little) familiarity with indexes in Taskcluster leads me to believe there isn't much of security concern unless we are taking about one project polluting another project, which isn't really related to the index format.

The `v3` format is specifically being proposed to prevent this from happening. `v2` only contains repo names, so if you have multiple repos with the same names (but different orgs) there's potential for pollution.

hwine:
> `v2` only contains repo names, so if you have multiple repos with the same names (but different orgs) there's potential for pollution.

Oh I may have misunderstood then. BTW, you should capture the `v4id` for repos -- both org & repo names can be changed, allowing for takeovers.

escapewindow:
> I _think_ we're getting close to consensus here. I think the two things left are:
> 
>     1. Making sure the `v3` index format is sensible, fleshed out enough, and solves the problem it's meant to. @hwine, @escapewindow - what do you think here?

I think so. I think this will have multiple parts:
- what does the v3 index look like? how will it fit our future use cases?
- are we migrating gecko, or just standalone taskgraph? if we're migrating, what does that look like? is there a watershed for e.g. mozregression?
- are we migrating all standalone taskgraph projects (e.g. mobile) or are we somehow limiting this to the Mozilla trust domain? if we're migrating, what does that look like?
- if we're supporting multiple index formats, what does that look like? how do we keep this from growing too complex, especially if there's a tool that needs to query multiple projects with different index formats?

>     2. Making sure the new LDAP group is doable and accepable. I'm pretty sure it's fine on our side (please speak up if you disagree), and I'm going to have a chat with IT about their side.

As mentioned elsewhere, I think we could have a similar process to the level 1 commit privileges agreement that allows us to know what github usernames should be enabled.

bhearsum:
> > `v2` only contains repo names, so if you have multiple repos with the same names (but different orgs) there's potential for pollution.
> 
> Oh I may have misunderstood then. BTW, you should capture the `v4id` for repos -- both org & repo names can be changed, allowing for takeovers.

This is not something I particularly want to include in an index, as they are meant to human readable. Is it a requirement?

escapewindow:
> I think it's pretty clear that a separate RFC makes sense for this. It's a big enough thing on its (more than I expected), and has a separate set of stakeholders. Do you want to block resolution of this on that being resolved? I could go either way.

I think I'd be happy here if we say, "we need a new index format, schema and specifics TBD, but our requirements here are that we don't have index collisions between repositories of the same name but different orgs" or something. It sounds like we may need to open the v3 index discussion to allow for standalone taskgraph and gecko taskgraph to live together, and other consumers may have use cases that help us decide what that thing looks like. Hopefully we can get this hashed out soonish, but we may want to take enough time to get a v3 index format that will stick for a while, and not have to resort to a v3.1 or v4 soon after.

I'm hoping it's clear that I'm raising concerns that will help improve the project, and I'm not trying to throw arbitrary obstacles in your way. I'd love for this to be a simple problem. Thank you for tackling this!

bhearsum:
> > I think it's pretty clear that a separate RFC makes sense for this. It's a big enough thing on its (more than I expected), and has a separate set of stakeholders. Do you want to block resolution of this on that being resolved? I could go either way.
> 
> I think I'd be happy here if we say, "we need a new index format, schema and specifics TBD, but our requirements here are that we don't have index collisions between repositories of the same name but different orgs" or something. It sounds like we may need to open the v3 index discussion to allow for standalone taskgraph and gecko taskgraph to live together, and other consumers may have use cases that help us decide what that thing looks like. Hopefully we can get this hashed out soonish, but we may want to take enough time to get a v3 index format that will stick for a while, and not have to resort to a v3.1 or v4 soon after.

OK, let's not block moving this along in that case, and I will make sure that I push along index discussions elsewhere in short order.

> I'm hoping it's clear that I'm raising concerns that will help improve the project, and I'm not trying to throw arbitrary obstacles in your way. I'd love for this to be a simple problem. Thank you for tackling this!

You're just pointing out obstacles that are already there, it's super helpful!

escapewindow:
Some thoughts I had last night on indexes:

Aiui, mshal spent multiple months and/or quarters refining and implementing the v2 index format for Gecko. This was a complex problem. We may be able to bug spelunk to determine his requirements and solutions, though we may have shoehorned additional use cases on top of Gecko v2 index since then.

However, we then also implemented another v2 index format for standalone taskgraph, and I suspect that the two v2 index formats are not fully compatible, which can cause issues when we use standalone taskgraph in Gecko. We may want to define the requirements here: finding decision tasks, cached tasks, toolchain + docker-image formats, treeherder, etc. as well as existing indexes for current standalone taskgraph projects. Plus, with the new Release for Mozilla RFCs, we want to avoid index collisions between projects of the same name but different organizations.

I suspect if we gather all of the above, we'll have a good idea of the requirements for a consolidated v3 index format. If we look at the audience involved in the v2 Gecko index format, we'll have part of the answer of "who is affected by an index format change", although we'll need to answer that question for standalone taskgraph projects as well. I suspect we'll want that shared and standardized v3 index format to solve all the above use cases. We probably want to send this proposal out to a wider audience, since this will affect downstream consumers.

However, I understand that while a consolidated v3 index format that solves all our combined use cases would be preferable, it's also a lot of work. I could also see proposing adding a v3 index format specifically for the `mozilla` trust domain, and dealing with the incompatible v2/v3 split across trust domains in build-decision, ci-config, ci-admin, standalone taskgraph, and any other tooling that queries indexes: shipit, CoT, ? If that is a preferred solution for the intermediate-term, we could do that, and plan to reach a consolidated v4 index format later down the road.

bhearsum:
Thanks for all of this, it's great to have some of the historical perspective here.

> Some thoughts I had last night on indexes:
> 
> Aiui, mshal spent multiple months and/or quarters refining and implementing the v2 index format for Gecko. This was a complex problem. We may be able to bug spelunk to determine his requirements and solutions, though we may have shoehorned additional use cases on top of Gecko v2 index since then.

Regardless of what we do here, looking into what we did for the last change to the index format seems like a good idea if we're proposing a new one :).

> However, I understand that while a consolidated v3 index format that solves all our combined use cases would be preferable, it's also a lot of work. I could also see proposing adding a v3 index format specifically for the `mozilla` trust domain, and dealing with the incompatible v2/v3 split across trust domains in build-decision, ci-config, ci-admin, standalone taskgraph, and any other tooling that queries indexes: shipit, CoT, ? If that is a preferred solution for the intermediate-term, we could do that, and plan to reach a consolidated v4 index format later down the road.

IMO, the only viable path forward in the medium-term is a new that's only used by either the `mozilla` trust domain, or everything that uses out-of-tree taskgraph. I can't find any justification for blocking this work on a new index format for Gecko (especially when it doesn't really provide clear benefit for it). I think we should make an effort to make the new format Gecko-compatible, but I don't think we need to go to the extent of consulting all of the Gecko stakeholders at this time (that's something we can do when we we're ready to have Gecko upgrade). I don't _think_ this is conflicting with anything you're saying, but please let me know if it is so we can resolve it.

There's also another option that we haven't given serious consideration, which is: don't change the index format at all. The only driver for it is avoiding collisions, and when I step back and look at it, the likelyhood of two repos having the exact same name is not terribly likely, so it may be worth thinking a bit more about the cost/benefit, and whether or not this is something we _must_ deal with now.

escapewindow:
> Regardless of what we do here, looking into what we did for the last change to the index format seems like a good idea if we're proposing a new one :).

+1 :)

> IMO, the only viable path forward in the medium-term is a new that's only used by either the `mozilla` trust domain, or everything that uses out-of-tree taskgraph. I can't find any justification for blocking this work on a new index format for Gecko (especially when it doesn't really provide clear benefit for it). I think we should make an effort to make the new format Gecko-compatible, but I don't think we need to go to the extent of consulting all of the Gecko stakeholders at this time (that's something we can do when we we're ready to have Gecko upgrade). I don't _think_ this is conflicting with anything you're saying, but please let me know if it is so we can resolve it.

Sure, I mentioned that's a possible stopgap v3, and then @ahal can do the v4 index compatibility work as a blocker to using standalone taskgraph in Gecko. Unless the standalone and gecko indexes are already compatible, which I think we're all hoping for.

> There's also another option that we haven't given serious consideration, which is: don't change the index format at all. The only driver for it is avoiding collisions, and when I step back and look at it, the likelyhood of two repos having the exact same name is not terribly likely, so it may be worth thinking a bit more about the cost/benefit, and whether or not this is something we _must_ deal with now.

This is interesting, and I gave it some thought.

We're helped here by the fact that the current proposal only targets level 1, and we've historically been ok with, say, Try tasks being capable of polluting other Try tasks, as long as Release tasks are secure.

Some thoughts here:

First, this would be easier to accept if, say, we knew that there were a `level-1` string in all the indexes that the `mozilla` trust domain has access to. I'm not certain that will be the case, e.g. [this block](https://hg.mozilla.org/ci/taskgraph/file/tip/src/taskgraph/transforms/task.py#l195) seems to have some index templates that don't refer to level at all. I'm a bit worried that later on down the line, we may want to add some level-3 indexes in this namespace, and those will be pollutable from the level-1 security projects: escalation of privilege. Some checks that verify that we don't have any `level > 1` stuff in the `mozilla` namespace may be good here.

Second, I'm not sure `mozilla` is a descriptive enough name now. Given that a) we know this is only for level-1, and b) we already use `project:mozilla:` as a prefix for some scopes (search for `project:mozilla:` in [this file](https://hg.mozilla.org/ci/ci-configuration/file/tip/grants.yml), for example), we may already have some level-3 scopes that may be affected by this new project. Maybe mozilla-level-1, mozilla-try, mozilla-ci, mozilla-auto-tc, github-auto, ?

Third, I'm not sure how uncommon name collisions will be. Maybe we don't care from a security perspective if `github.com/org1/persona` can pollute indexes from `github.com/org2/persona`, because it's level 1. However, from a usability standpoint, going to a `latest` index for your build or test and going to a task from another repo may end up wasting time and causing frustration or confusion. Even if we think that we will rarely have a `github.com/org{1,2}/persona` collision for two completely different projects, Github forks will automatically have the same project name as the original repo, and it's not far fetched to think that someone will want to enable Taskcluster on their fork of a Taskcluster-enabled original repo. If both are in the `mozilla` trust domain, then the fork will pollute the indexes of the main repo, and someone may think that they're download the latest bits from the main repo, but they're downloading the bits from the fork.

escapewindow:
I'll note that this collision problem predates this RFC: When enabling the staging repos for fenix, android-components, xpi-manifest, and adhoc-signing, I intentionally added a `staging-` prefix to these forks because otherwise the staging tasks would pollute the production index namespaces.

However, the `mozilla` trust domain will remove the human reviewer and automate adding these, so we accelerate the process with no human saying "um, we already have 5 `persona` repos, and that's causing a problem, so maybe a different name"? I gave some thought to the site or integration somehow checking a master list and saying "you can't install me on your `org2/persona` repo because there's already an `org1/persona` repo", but then it would have to catch repo renames and deleted repos with old indexes lying around and I'm not sure how complex we want our workaround solution to be.

escapewindow:
Another thought, sorry: we could limit the `mozilla` trust domain to a single org, no user repos, no 2nd org, which prevents name collisions. This reduces its functionality in that it's useful to enable TC on your fork to test out automation changes before creating a PR, but this could be a solution.

ahal:
For the collision problem couldn't we make sure the org is part of the
index? GitHub projects are always identified by `<org>/<repo>` everywhere
that I've seen.

escapewindow:
> For the collision problem couldn't we make sure the org is part of the index? GitHub projects are always identified by `<org>/<repo>` everywhere that I've seen.

Yes :) This echoes what I said [in April](https://github.com/mozilla-releng/releng-rfcs/pull/33#discussion_r612751707) :)

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,78 @@
...
+We will need hardware for 3 different pools, which will be allocated as noted below:
+* 2 machines for macOS signing, running allocated from the existing production Firefox pool
+* 3 machines for macOS builds, allocated from TBD
+* 3 machines for macOS tests, allocated from TBD
+
+When additional workers are needed in the future, they will be allocated from TBD.
+
+## `v3` Taskcluster index format
+
+The current `v2` index format only includes repository names as an identifier (not user or organization). This is generally not an issue for any of current trust domains (because they generally only support one project), but in this new pool where we have N projects, it introduces the potential for collisions or pollution between them. To ensure this isn't an issue we will introduce a new `v3` index format that includes the repository location in its path as well - including both domain and path. Examples include:
```
ahal, line 59: The mobile repos define a `trust_project` in their project.yml:
https://hg.mozilla.org/ci/ci-configuration/file/73a8d7921adb15a8ca3b726225655fc9114524a7/projects.yml#l485

This is then used in the index route. I think we could simply define this as `org/repo` for any projects that use the `mozilla` trust domain instead of a creating a format?

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,78 @@
...
+* `index.mozilla.v3.github.com.mozilla-mobile.mozilla-vpn-client.branch.main.latest.taskgraph.decision`
+* `index.gecko.v3.hg.mozilla.org.releases.mozilla-beta.latest.firefox.decision`
+
+This will require changes to a few things to support the new format:
+* build-decision
```
ahal, line 64: What build-decision changes were you anticipating? I'd guess any projects using this will be on Github.

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,78 @@
...
+We will create a new trust domain and workers that are generally available for Mozilla employees and trusted volunteers to use. Specifically:
+
+* A new Trust Domain (`mozilla`) that is not tied to a specific project or product
+* New Workers for builds on Linux, macOS 10.15, and Windows Server 2012
+    * A RelEng maintained Docker image will be provided for Linux
+    * These will be created under a new `mozilla-1` provisioner
+* New Workers for tests on Linux, macOS 10.15, and Windows 10
+    * A RelEng maintained Docker image will be provided for Linux
+    * These will be created under a new `mozilla-t` provisioner
+* New Scriptworker instances for signing and mac-signing
```
ahal, line 24: I'm a bit on the fence about this. On the one hand, I realize that the intent here is to make it easier for us to stand things up, especially for things like mac signers. On the other, maybe when a project reaches the point of needing scriptworkers, then that is a signal that it is time for them to graduate to their own trust domain.

rfcs/0033-shared-trust-domain.md
```diff
@@ -0,0 +1,78 @@
...
+	* These will be created under the existing `scriptworker-k8s` and `scriptworker-prov-v1` provisioners
+	* Workers will be prefixed with `mozilla-t-`
+    * mac-signing will run 10.14, like our other mac-signing workers (there's no known reason to upgrade)
+
+Notably, we are only concerned with level 1 workers at this time, which means we can ignore things like scriptworkers that are only used when shipping. Level 3 workers will be dealt with at a later stage, and most likely will not use a shared trust domain or workers across projects.
```
ahal, line 29: As a sort of follow-on to my previous comment, what if instead of disallowing L3 on the shared pools, we instead disallow scriptworkers? Then once the project is a bit more mature, we would graduate them to their own trust domain and add the scriptworker tasks.

This way they could get started with Taskcluster right away, and have a proper PR == L1 and push == L3 setup. Security wise, the setup wouldn't be ideal.. but it would still be better than GH Actions, or doing signing on personal dev machines.

It kind of reminds me of the autonomous driving quote. It doesn't have to be perfect, it just has to be better than manual.

rfcs/0033-shared-trust-domain.md
ahal: Also, if we're only supporting L1 scriptworkers on this pool, then is it really saving us any work? We'll still need to wrangle mac signers for the L3 pool anyway.

rfcs/0033-shared-trust-domain.md
bhearsum: Sounds fine to me - good idea!

rfcs/0033-shared-trust-domain.md
bhearsum: I...I have no idea at this point. Maybe I was thinking that some new projects might not be on Git. We can probably remove/ignore this.

rfcs/0033-shared-trust-domain.md
bhearsum: I think this loses a fair amount of value if we can't have signing as part of this. We'd have to do one of the following pretty quickly:
- Manually sign (avoiding this was part of the movitation behind this idea)
- Move projects very quickly from the shared pool to their own exclusive one
- Have automation to sign outside of scriptworkers (eg: a task that signs through autograph-edge). I don't even know if SecEng would allow this though :).

A potentially crazier idea could be to allow most tasks to run in the shared pool, but somehow put signing in its own pool (which means tasks in the same graph would be have two trust domains...which would probably means CoT changes). This would at least mean that signing would be purely adding to existing things, rather than getting blocked on a migration to a different pool.

rfcs/0033-shared-trust-domain.md
bhearsum: > As a sort of follow-on to my previous comment, what if instead of disallowing L3 on the shared pools, we instead disallow scriptworkers?

I think this is a great idea, independent of what we do with any other part of this! I can't think of any downsides to separating PRs and pushes.

bhearsum:
@ahal - is this RFC still relevant given your recent work on this?

ahal:
Hmm, yeah I guess I should have merged this first. There might still be some open issues around non-level 1 / scriptworker pools.. but yeah, it's mostly implemented already.

PR #35, opened 2021-04-22
RFC 35: Easy hg/github → hg.mozilla.org syncing

bhearsum:
[rendered](https://github.com/bhearsum/releng-rfcs/blob/github-syncing/rfcs/0035-simple-repo-syncing.md)

### DIFF
--- a/rfcs/0035-simple-repo-syncing.md
+++ b/rfcs/0035-simple-repo-syncing.md
@@ -1,0 +1,72 @@
+# RFC 35 - Easy hg/github → hg.mozilla.org syncing
+* Comments: [#35](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/35)
+* Proposed by: @bhearsum
+
+# Summary
+
+Support one-way syncing from existing hg or Github repos into Phabricator-enabled repositories.
+
+## Motivation
+
+A handful of projects (mostly notably PDF.js and NSS) regularly need to sync code into Gecko. In all cases, this is currently done by hand on laptops, making it annoying, error-prone, and low bus factor. Supporting this as a first class operation would make things better & easier for existing projects, and allow others to do so much more easily, enabling other use cases.
+
+## Out of Scope
+
+This RFC is intentionally tightly scoped to the currently known use cases. In particular, we are specifically _not_ supporting 2-way syncs or syncing things _into_ a Git or Github repository as part of this. These things may be considered in the future, but are explicitly out of scope for now.
+
+# Details
+
+Because this RFC is scoped to only support Phabricator-enabled repositories we can ignore some of the usual security concerns (because we won't have high-value credentials, just a Phabricator token). This also means we don't need to use something as complex as a scriptworker -- docker-worker should be perfectly fine for this use case. Most of the implemention work will be glueing existing things together. At a high level, we need a Task that:
+
+* Takes the artifact from an upstream Task
+* Unpacks it
+* Clones the target repository
+* Copies all or some of the contents to a target repository
+* Runs `moz-phab` to create a Phabricator revision
+
+Choosing to take an artifact from an upstream Task rather than simply copy files from a source repository clone is notable, as it provides the flexibility for projects to run build or other preprocessing steps before syncing is completed (which is a known use case for both PDF.js and NSS). If uses cases for more simple copies present themselves in the future, we may add support for the simpler operation as well - but it is not necessary at this time.
+
+## Taskgraph
+
+### New Docker Image
+
+A new Docker image will be created that provides Mercurial (including any necessary tools to do fast clones), `moz-phab`, and a simple helper script that wraps the process described above. (This helper script primarily exists to ensure we never run into command line length limits.)
+
+### Kind & Transforms
+
+A new `kind` will be created that can create sync jobs. Sensible defaults will be used for descriptions, treeherder symbols, etc. It will run on docker-worker with the aforementioned image.
+
+This kind will need a new transform that will take information provided by a project repository (see below), and use it to construct the right command to execute.
+
+## Project Repositories
+
+Will need to provide:
+
+* Target repository URL
+* Upstream task name and artifact name
+* Source -> Target mapping, relatively to artifact root & target repository root, ie:
+
+```yaml::
+syncMappings:
+    targetRepository: https://hg.mozilla.org/mozilla-central
+    sourceArtifacts:
+        - {buildTask}/buildArtifact.tar.bz2:
+            src/: toolkit/components/pdfjs/content/build/
+            web/: toolkit/components/pdfjs/content/web/
+```
+
+This information will be provided in the simplified Taskcluster config described by RFC XXX. (This needs updating after that RFC is opened.)
+
+# Open Questions
+
+* Does this enable single actor attacks? Ie: a privileged person in the PDF.js repo gets automation to create a Phab revision, and then reviews it themselves and lands it?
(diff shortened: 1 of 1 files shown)


rfcs/0035-simple-repo-syncing.md
```diff
@@ -0,0 +1,72 @@
...
+            src/: toolkit/components/pdfjs/content/build/
+            web/: toolkit/components/pdfjs/content/web/
+```
+
+This information will be provided in the simplified Taskcluster config described by RFC XXX. (This needs updating after that RFC is opened.)
+
+# Open Questions
+
+* Does this enable single actor attacks? Ie: a privileged person in the PDF.js repo gets automation to create a Phab revision, and then reviews it themselves and lands it?
+    * I think people can already land without review, so maybe it's effectively no different?
```
escapewindow, line 63: I think so, as written now, before my comments:

- someone extracts the phab token, somehow
- someone creates a phab patch that looks like a valid phab patch, but contains malicious contents
- waits for someone to review it, or they review it themselves
- no tests run, and we've pwned central

escapewindow: `I think people can already land without review, so maybe it's effectively no different?`

If they do so, it probably looks like they landed their own patch with an `r=me` and that patch is likely worth some inspection by other developers.

If they follow this approach, it looks like they reviewed a valid vendoring patch, and these patches can be large enough that finding the badness can be a needle in a haystack without the verification task.

escapewindow (changes requested):
Marking `request changes` because I think we can't proceed without changes.

rfcs/0035-simple-repo-syncing.md
```diff
@@ -0,0 +1,72 @@
...
+
+# Details
+
+Because this RFC is scoped to only support Phabricator-enabled repositories we can ignore some of the usual security concerns (because we won't have high-value credentials, just a Phabricator token). This also means we don't need to use something as complex as a scriptworker -- docker-worker should be perfectly fine for this use case. Most of the implemention work will be glueing existing things together. At a high level, we need a Task that:
+
+* Takes the artifact from an upstream Task
+* Unpacks it
+* Clones the target repository
+* Copies all or some of the contents to a target repository
+* Runs `moz-phab` to create a Phabricator revision
```
hwine, line 25: A key safety feature here is that none of the just-modified code can be executed during the landing process. It's not clear to me if `moz-phab` enforces that rule.

Related, I'd like to see some hard coding as to what parts of the repository are allowed to be updated via this process. This type of process should not be able to update any of the build code (including taskgraphs), nor modify any file that will be passed to `eval()` type functions. 

If these precautions are possible, please doc as requirements. If not possible - how else can we protect against that?

rfcs/0035-simple-repo-syncing.md
```diff
@@ -0,0 +1,72 @@
...
+            web/: toolkit/components/pdfjs/content/web/
+```
+
+This information will be provided in the simplified Taskcluster config described by RFC XXX. (This needs updating after that RFC is opened.)
+
+# Open Questions
+
+* Does this enable single actor attacks? Ie: a privileged person in the PDF.js repo gets automation to create a Phab revision, and then reviews it themselves and lands it?
+    * I think people can already land without review, so maybe it's effectively no different?
+
```
hwine, line 64: How will use of this feature be restricted? It's one thing to have it used in vetted situations, it's quite another to allow general usage. The "only in this subtree" restriction suggested above is one level of protection, but are any more planned?

I'm assuming this would only run on level 3 builders "for real", but I don't know how that impacts testing, etc.

rfcs/0035-simple-repo-syncing.md
```diff
@@ -0,0 +1,72 @@
...
+A handful of projects (mostly notably PDF.js and NSS) regularly need to sync code into Gecko. In all cases, this is currently done by hand on laptops, making it annoying, error-prone, and low bus factor. Supporting this as a first class operation would make things better & easier for existing projects, and allow others to do so much more easily, enabling other use cases.
+
+## Out of Scope
+
+This RFC is intentionally tightly scoped to the currently known use cases. In particular, we are specifically _not_ supporting 2-way syncs or syncing things _into_ a Git or Github repository as part of this. These things may be considered in the future, but are explicitly out of scope for now.
+
+# Details
+
+Because this RFC is scoped to only support Phabricator-enabled repositories we can ignore some of the usual security concerns (because we won't have high-value credentials, just a Phabricator token). This also means we don't need to use something as complex as a scriptworker -- docker-worker should be perfectly fine for this use case. Most of the implemention work will be glueing existing things together. At a high level, we need a Task that:
```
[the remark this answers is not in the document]
bhearsum: This is a really good point; I somehow managed to forget about updatebot when writing this :upside_down_face:

rfcs/0035-simple-repo-syncing.md
```diff
@@ -0,0 +1,72 @@
...
+
+* Takes the artifact from an upstream Task
+* Unpacks it
+* Clones the target repository
+* Copies all or some of the contents to a target repository
+* Runs `moz-phab` to create a Phabricator revision
+
+Choosing to take an artifact from an upstream Task rather than simply copy files from a source repository clone is notable, as it provides the flexibility for projects to run build or other preprocessing steps before syncing is completed (which is a known use case for both PDF.js and NSS). If uses cases for more simple copies present themselves in the future, we may add support for the simpler operation as well - but it is not necessary at this time.
```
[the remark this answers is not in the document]
bhearsum: Is this really a requirement, or a nice to have? The alternative we're comparing to is people doing similar things on their laptops.

rfcs/0035-simple-repo-syncing.md
bhearsum: It's a good point that the look of someone landing their own patch is very different here. I definitely agree that it's less likely to be noticed at something alarming.

rfcs/0035-simple-repo-syncing.md
bhearsum: > This type of process should not be able to update any of the build code (including taskgraphs), nor modify any file that will be passed to eval() type functions.

I'm not sure this is practical, if I'm understanding correctly. As an example, NSS imports into mozilla-central, including files like https://hg.mozilla.org/mozilla-central/file/tip/security/nss/build.sh, which can obviously eval and run arbitrary code. Is this something you would want to be denied landing through this system?

hwine: >  Is this something you would want to be denied landing through this system?

Yes, unless we can think of sufficient "other layers" to mitigate the risk. Combined with the SaC & approve-own-patch issue, I'm feeling less comfortable. :/

You may want to ask for an early RRA/consultation -- I'm unsure how the "risks of new approach" compare to "risks of current system"

rfcs/0035-simple-repo-syncing.md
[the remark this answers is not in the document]
escapewindow: I think if we're doing it properly, not allowing for a single bad actor to pwn m-c is preferred.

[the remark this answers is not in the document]
escapewindow: I think if we do updatebot or similar from the gecko trust domain, cloning the source repo and running a script, then we can use CoT to verify the artifact is from a trusted tree/worker, and we can worry a bit less about reproducibility.

bhearsum:
This is irrelevant given our move to Github as the sole supported. VCS.

PR #36, opened 2021-04-28
RFC 36: Enhance Taskgraph to support the common build → test workflow

bhearsum:
[rendered](https://github.com/bhearsum/releng-rfcs/blob/taskgraph-workflow/rfcs/0036-taskgraph-workflow.md)

### DIFF
--- a/rfcs/0036-taskgraph-workflow.md
+++ b/rfcs/0036-taskgraph-workflow.md
@@ -1,0 +1,258 @@
+# RFC 36 - Enhance Taskgraph to support the common build → test workflow
+* Comments: [#36](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/36)
+* Proposed by: @bhearsum
+
+# Summary
+
+Enchance Taskgraph to support the common build → test workflow without requiring custom code in a project repository.
+
+## Motivation
+
+One of the barriers to entry for using Taskcluster is getting usable configs and support files ready and working properly. This is needed even when a project just needs simple builds and tests, and can be both intimidating and complex. Adding support for this common workflow in Taskgraph will make it much quicker and easier to onboard with projects.
+
+# Details
+
+## Division of Responsibilities
+
+A conceit of this RFC is that while no custom _code_ should be needed to do simple builds and tests, we will still require some minimal amount of _config_. Effectively this will shift most of the responsibility and complexity for simple builds and tests from project repos to Taskgraph itself.
+
+Project repositories will only need to provide simple information that should be mostly obvious to their owners:
+* A list of build & test platforms, as well as the following for each:
+    * Command used to build or test
+    * Worker type and docker image name (when applicable)
+    * Artifacts to upload
+    * Dependent builds or tests, and artifacts required from them
+
+## New Taskcluster Config
+
+A new, simplified, Taskcluster config will be created. It will be located in `.taskcluster/config.yml` and contain:
+* A reference to another config file, which will contain all of the things normally found in a `.taskcluster.yml`
+* A list of tasks using a new, simplified schema
+
+Here's an example:
+
+```yaml::
+version: 1
+baseConfig: https://some/releng/owned/thing/taskcluster.yml-v1
+extraTransforms:
+    - treeherder
+tasks:
+    build-linux64:
+        workerType: b-linux
+        docker-image: build
+        command: make build
+        produces:
+            linux64.tar.bz2: build/linux64.tar.bz2
+            tests.tar.bz2: build/tests.tar.bz2
+        notify:
+            slack:
+                - CKM7DLL67 # #myproject
+            matrix:
+                - !jmMnZXHjnLBJCfUIWh:mozilla.org # #myproject
+            
+    test-linux64:
+        workerType: t-linux
+        docker-image: ubuntu1804
+        consumes:
+            build-linux64:
+                - linux64.tar.bz2
+                - tests.tar.bz2
+        command: tar -jvxf tests.tar.bz2 && run_tests.py --build linux64.tar.bz2
+```
+
+This config is explicitly trying to be as simple as possible by doing a few things:
+* Assuming that tasks are always a command running on a given worker/image
+* Not structurally separating task metadata and payload data
+* Explicitly _not_ supporting JSON-e. Complex use cases should be moved to more traditional ways of defining tasks (see below).
+
+In some ways this is an even more simplified version of the `run-task` job type used in the Gecko tree.
+
+The full schema for version 1 is as follows:
+```yaml::
+required:
+    - version
+    - baseConfig
+    - tasks
+
+properties:
+    version:
+        type: integer
+        description: Config Schema Version
+        default: 1
+        enum:
+            - 1
+
+    baseConfig:
+        type: string
+        description: A link to the .taskcluster.yml to use when generating tasks for this repository.
+        default: https://some/releng/owned/thing/taskcluster.yml-v1
+
+    extraTransforms:
+        type: array
+        description: A list of extra transforms to apply to the Tasks in this config
+        items:
+            type: string
+            description: Transform name
+            examples:
+                - treeherder
+
+    tasks:
+        type: object
+        description: A set of tasks tasks to run in response to pushes and pull requests, keyed by task name
+        patternProperties:
+            ^.*$:
+                type: object
+                required:
+                    - command
+                    - workerType
+                properties:
+                    produces:
+                        type: object
+                        description: Artifacts produced by the task, keyed by what they should be called
+                        patternProperties:
+                            ^.*$:
+                                type: string
+                                description: Path to the artifact, relative to the root of the build directory
+
+                    command:
+                        type: string
+                        description: Command used to run the Task. Complex commands should generally be avoided, and moved into a checked-in script.
+
+                    env:
+                        type: object
+                        description: Environment variables to set for the Task, keyed by variable name
+                        patternProperties:
+                            ^.*$:
+                                type: string
+                                description: Variable value
+
+                    notify:
+                        type: object
+                        properties:
+                            matrix:
+                                type: array
+                                items:
+                                    type: string
+                                    description: Matrix channel ID to send a notification to on task failures
+
+                            slack:
+                                type: array
+                                items:
+                                    type: string
+                                    description: Slack channel ID to send a notification to on task failures
+
+                    dockerImage:
+                        type: string
+                        description: Docker image to run the Task within. Only valid (and required) for Linux workers.
+
+                    consumes:
+                        type: object
+                        description: Artifacts from other Tasks that this Task consumes, keyed by upstream Task name.
+                        patternProperties:
+                            ^.*$:
+                                type: array
+                                items:
+                                    type: string
+                                    description: Artifacts to download from the named upstream Task
+
+                    workerType:
+                        type: string
+                        description: Worker to run this Task on
+                        enum:
+                            - b-linux
+                            - b-macos11.0
+                            - b-windows2012
+                            - t-linux
+                            - t-macosx11.0
+                            - t-windows10
+```
+
+## Adding complex Taskcluster configuration
+
+It will not be possible (nor desireable) to implement all possible tasks within the simplified config described above. In particular, Tasks that need custom loaders or transforms will need to be defined in the traditional way - with a custom kind that fully describes the Task, loader, and any necessary transforms. Taskgraph will be able to handle these types of Tasks referencing those found in the simplified config, and vice versa. This allows simple Tasks to be used as much as possible, and to opt-in to more complex Tasks only where needed. Here's an example of how this would work:
+
+`.taskcluster/config.yml`:
+```yaml::
+baseConfig: https://some/releng/owned/thing/taskcluster.yml-v1
+extraTransforms:
+    - treeherder
+tasks:
+    build-macos:
+        workerType: b-macos
+        command: xcode build
+        produces:
+            macos-unsigned.dmg: build/macos-unsigned.dmg
+            tests.tar.bz2: build/tests.tar.bz2
+    test-macos:
+        workerType: t-macos
+        consumes:
+            # This is a reference to the task above
+            build-macos:
+                - tests.tar.bz2
+            # This is a reference to the task in `kinds/sign/kind.yml`
+            sign-macos:
+                - macos-signed.dmg
+        command: tar -jvxf tests.tar.bz2 && run_tests.py --build macos-signed.dmg
+```
+
+`.taskcluster/kinds/sign/kind.yml`:
+```yaml::
+loader: taskgraph.loader.transform:loader
+
+transforms:
+    - taskgraph.transforms.signing:transforms
+
+jobs:
+    sign-macos:
+        description: sign macos build
+        name: signing
+        worker-type: scriptworker/mozilla-1-iscript
+        worker:
+            implementation: iscript
+            cert: dep
+            upstreamArtifacts:
+                # This is a reference to the task in `config.yml`
+                build-macos:
+                    macos-unsigned.dmg:
+                        formats:
+                            - mac_sign
+```
+
+## Taskcluster-Github changes
+
+Taskcluster-Github will need to be updated to support the new `.taskcluster/config.yml` config. When it receives an event, and that file is present in the repository, it will:
+* Dereference `baseConfig` to find the `.taskcluster.yml` 
+* Process the `.taskcluster.yml` as usual (in most cases this will result in a decision task being created)
+
+A new context variable, `simple_config`, will be introduced. When a `.taskcluster/config.yml` file is present in the repository this variable will be populated with an URL to it.
+
+## Taskgraph changes
+
+In order to support transforming simple configs into actual tasks `taskgraph` will need to provide a number of things it currently does not. Things that aren't Mozilla-specific will go in core `taskgraph`, while things that are will go in a new `mozilla_taskgraph` module. This will be similar to what we already do we modules such as `fenix_taskgraph`, except that it will be a place for storing things that are generally useful to Mozilla projects - not just a single one.
+
+In core `taskgraph`, we will provide:
+* Docker images for builds and tests on Linux (detailed in [RFC #33](https://github.com/mozilla-releng/releng-rfcs/pull/33))
+* A .taskcluster.yml file that supports `push`, `pull-request`, and `action` `tasks_for`
+* A `pr-complete` kind (like we use in many existing repos, to have a simple way to know when a Task Group has completed)
+* A `simple_task` kind, loader, and transform that can translate tasks from `.taskcluster/config.yml` to real Tasks
+    * The `simple_task` loader will work similar to the `jobs-from` mechanism used in Gecko, but source Tasks from `.taskcluster/config.yml`
+    * Including sensible defaults for maximum runtimes & task priorities
+
+In `mozilla_taskgraph`, we will provide:
+* Worker definitions for available build and test machines (detailed in [RFC #33](https://github.com/mozilla-releng/releng-rfcs/pull/33))
+* A `mozilla_task` transform that adds Mozilla-specific configuration and defaults to simple tasks (most notably, `trust-domain` and Treeherder metadata)
+
+Notably, there are is no special handling for `build` and `test` Tasks - both are handled by the `simple_task` kind. Because the types of things that would normally be different between builds and tests will be defined in project repos (docker images, workers, dependencies, etc.), there is no need to have separate kinds in taskgraph.
+
+Because we are separating out Mozilla-specific things into a separate module, we need a way to register that module before any decision task logic is run (to ensure everything inside of it is accessible when generating Tasks). This will be accomplished by adding support for a new `extra-modules` parameter to the decision task command, and the RelEng maintained `.taskcluster.yml` files will ensure `mozilla_taskgraph` is available, and passed to the decision task.
+
+# Open Questions
+
+# Implementation
+
+<once the RFC is decided, these links will provide readers a way to track the
+implementation through to completion>
+
+* <link to tracker bug, issue, etc.>
+* <...>
+


rfcs/0036-taskgraph-workflow.md
```diff
@@ -0,0 +1,258 @@
...
+                            - b-macos11.0
+                            - b-windows2012
+                            - t-linux
+                            - t-macosx11.0
+                            - t-windows10
+```
+
+## Adding complex Taskcluster configuration
+
+It will not be possible (nor desireable) to implement all possible tasks within the simplified config described above. In particular, Tasks that need custom loaders or transforms will need to be defined in the traditional way - with a custom kind that fully describes the Task, loader, and any necessary transforms. Taskgraph will be able to handle these types of Tasks referencing those found in the simplified config, and vice versa. This allows simple Tasks to be used as much as possible, and to opt-in to more complex Tasks only where needed. Here's an example of how this would work:
```
djmitche, line 172: This kind of graceful fallback is 🥇

rfcs/0036-taskgraph-workflow.md
```diff
@@ -0,0 +1,258 @@
...
+                        type: object
+                        description: Artifacts from other Tasks that this Task consumes, keyed by upstream Task name.
+                        patternProperties:
+                            ^.*$:
+                                type: array
+                                items:
+                                    type: string
+                                    description: Artifacts to download from the named upstream Task
+
+                    workerType:
```
djmitche, line 158: We're getting away from the term `workerType`, but I see that there's no `provisionerId` anywhere here, so maybe this is really a `workerPoolSuffix` or something, with the prefix (=`provisionerId`) determined by the repo?

rfcs/0036-taskgraph-workflow.md
```diff
@@ -0,0 +1,258 @@
...
+                    macos-unsigned.dmg:
+                        formats:
+                            - mac_sign
+```
+
+## Taskcluster-Github changes
+
+Taskcluster-Github will need to be updated to support the new `.taskcluster/config.yml` config. When it receives an event, and that file is present in the repository, it will:
+* Dereference `baseConfig` to find the `.taskcluster.yml` 
+* Process the `.taskcluster.yml` as usual (in most cases this will result in a decision task being created)
```
djmitche, line 225: I like the proposal to have Taskcluster-Github do this kind of operation, and @imbstack has had similar thoughts in the past (and may have some refinements to add to this idea).

I think renaming `.taskcluster.yml` -> `.taskcluster/config.yml` is probably a separate concern.  This proposal could be accomplished with the filename `.taskcluster.yml`, I think.

rfcs/0036-taskgraph-workflow.md
```diff
@@ -0,0 +1,258 @@
...
+
+A new context variable, `simple_config`, will be introduced. When a `.taskcluster/config.yml` file is present in the repository this variable will be populated with an URL to it.
+
+## Taskgraph changes
+
+In order to support transforming simple configs into actual tasks `taskgraph` will need to provide a number of things it currently does not. Things that aren't Mozilla-specific will go in core `taskgraph`, while things that are will go in a new `mozilla_taskgraph` module. This will be similar to what we already do we modules such as `fenix_taskgraph`, except that it will be a place for storing things that are generally useful to Mozilla projects - not just a single one.
```
djmitche, line 231: I like this separation of names.  There may be a few other places where "taskgraph" would be a better word than "taskcluster" to avoid confusion, but I think you've covered them.

rfcs/0036-taskgraph-workflow.md
bhearsum: I'm not exactly sure what the right term for it is, but I'm thinking this is just a simple string that maps to entries defined somewhere in `mozilla_taskgraph`. Something like the entries we already use in files like https://searchfox.org/mozilla-central/source/taskcluster/ci/config.yml#426. So maybe `workerAlias` is a better name? I'm not too fussy on what it's called though.

rfcs/0036-taskgraph-workflow.md
bhearsum: I'm good with that. I mostly went with `.taskcluster/config.yml` because I thought it would avoid confusion. I'm happy to defer to you folks on this though.

rfcs/0036-taskgraph-workflow.md
bhearsum: Other places in this RFC, or somewhere else?

rfcs/0036-taskgraph-workflow.md
djmitche: `workerType` might be fine, as long as it's defined

rfcs/0036-taskgraph-workflow.md
djmitche: I'm not sure.  The use of a `taskcluster` top-level directory in repositories is an example.  I'm not saying to change that in this RFC, just that it has caused confusion and where possible it'd be nice to avoid it :)

rfcs/0036-taskgraph-workflow.md
```diff
@@ -0,0 +1,258 @@
...
+* A reference to another config file, which will contain all of the things normally found in a `.taskcluster.yml`
+* A list of tasks using a new, simplified schema
+
+Here's an example:
+
+```yaml::
+version: 1
+baseConfig: https://some/releng/owned/thing/taskcluster.yml-v1
+extraTransforms:
+    - treeherder
```
ahal, line 38: Where will these extra transforms be defined? These `treeherder` transforms would likely go in `mozilla_taskgraph`, but there might be opt-in transforms in the core `taskgraph` as well. Maybe each taskgraph module (`taskgraph`, `mozilla_taskgraph`, etc) could have its own `baseConfig` and which transforms are available would depend on the `baseConfig` a project decides to use.

rfcs/0036-taskgraph-workflow.md
ahal: Maybe a bit bikeshedd-y for this stage, but I kind of like having it under a hidden dir (either `.taskcluster` or `.taskgraph` or whatever). That way no matter whether you are using the simple config or the complex case, you always know to look in the same place.

I agree this change is somewhat tangential to this particular RFC, though I do think it would tie in to the broader "Release for Mozilla" proposal.

bhearsum: This is more tangential than I thought before I published this :). I wonder if this warrants a separate RFC for the tc-gh changes we'll need to support this?

rfcs/0036-taskgraph-workflow.md
bhearsum: Just to clarify, I'm assuming you're suggesting that, for example, we'd have one baseConfig that supports `taskgraph` only, and one that supports `taskgraph` and  `mozilla_taskgraph`? (In practical terms, they would be virtually the same, except the latter would find/install `mozilla_taskgraph` and tell `taskgraph decision` about it.)

If that's what you mean, I think that makes sense.

ahal: Yeah, something like that. Or maybe `baseConfigs` could include other `baseConfigs`.. I don't know. This is also a problem we could worry about later. Initially I'm sure everything will be using the "mozilla centric" baseConfig. But if we're going to take the step of offering a Mozilla-free `taskgraph` module, we'll likely want to at least think about this.

rfcs/0036-taskgraph-workflow.md
djmitche: I think the "main" tc-github changes can be included here.

I think adding an option to put the in-repo file from which that service reads at *either* `.taskcluster.yml` or `.taskcluster/config.yml` is also a fine idea (although others may disagree), and could probably land as just a PR to the service without an RFC.

What I'd like to avoid is that `.taskcluster.yml` acts one way, and `.taskcluster/config.yml` acts another way.

bhearsum: Right, OK - so I think this may not be written clearly enough, because this is exactly why I chose this new file for the simple config. Whether it's `.taskcluster.yml` or some new place, we still need a file in the existing format to handle decision task creation. The simple config is a _different_ config that gets passed to the decision task. (Not confusing at all !)

djmitche: Huh, I can't see that reading in the text this thread is attached to.  Absent knowledge of taskgraph, the documentation for tc-github would read a bit strangely, then: there are two config files, but tc-github doesn't care what's in one of them, and just provides a URL to it in context when rendering the other one.

I think there are a few pieces here which can be combined in a more natural fashion:

* renaming the config file (without changing its meaning)
* `baseConfig` property containing a URL saying "use this as well to define the task(s) for GH events"
* providing arbitrary-shaped additional configuration to the decision task

What if we define `baseConfig` as a feature (modulo suggestions from @imbstack, who may need an email ping to get him to look at this), and then address the last point by allowing an `extra` property (or a better name) which is passed as `extra` in the JSON-e context?  So `.taskcluster.yml`:

```yaml
version: 1
baseConfig: https://raw.githubusercontent.com/mozilla-releng/taskgraph/v1.2.3/baseConfig.yml
extra:
  extraTransforms:
    - treeherder
  tasks:  # or maybe `jobs`?
    build-linux64: ...
```

The `baseConfig` for taskgraph could then pass `env: {TCYML_EXTRA: {$json: {$eval: extra}}}` to pass that along to taskgraph, but other base configs might use some values from it directly.

If `baseConfig` is defined to do some kind of merge of the given `.taskcluster.yml` and the base config, then `.taskcluster.yml` could even include its own `tasks` that is appended to that in the base config.

ahal: I was thinking about this a bit last night, and I think having two separate config files is going to be something we regret down the line. Supporting a single config file that can handle both the simple and complex use cases feels like it will produce less confusion, while potentially also helping the use case where users would want some "simple tasks" as well as more complex kinds.

What if `baseConfig` could be any of:

1. A URL to a `baseConfig`
2. A file path to a `baseConfig`
3. An object (contents that were in the old-style `.taskcluster.yml` could just go here)
4. Absent. In this case we'd use the default `baseConfig` that ships with the core `taskgraph` module for the specified `version`.
5. If we wanted to go crazy, we could optionally make it a list where each item in the list could be any of 1-3, where latter `baseConfigs` override values from earlier `baseConfigs` (I don't think we want to go crazy though)

In this way there would only need to be one config file and it could grow organically from simple -> complex without the need of a hard switch over.

rfcs/0036-taskgraph-workflow.md
```diff
@@ -0,0 +1,258 @@
...
+tasks:
```
ahal, line 39: I think it would be worth also supporting something similar to m-c's `jobs-from`:
https://searchfox.org/mozilla-central/source/taskcluster/ci/test/kind.yml#27

That way simple tasks could logically be separated by type:
```
tasks-from:
    - build.yml
    - test.yml
    - lint.yml
```

The task format could be the same.

rfcs/0036-taskgraph-workflow.md
bhearsum: Let me try to make sure I understand correctly. Let's assume we're staying with `.taskcluster.yml` for now for simplicity....we just have one config in the repo (above). When an event comes in, tc-gh will:
* Dereference `baseConfig`
* Perform JSON-e substitution on the `baseConfig`, passing all of the usual variables, plus the `extra` block from `.taskcluster.yml` to it
* Create any tasks specified by the processed `baseConfig` (probably a decision task).

And in the specific case we're talking about here, we'd most likely end up passing the `extra` block to the decision task through an environment variable?

I think this could make sense; although @ahal just added a comment below about potentially supporting a `jobs-from` type of mechanism here which may make that trickier. (I think we'd have to load those files over http instead of through the environment, unless we add special handling for that in tc-gh as well.)

My other question is whether or not we'd still support JSON-e in this new version of `.taskcluster.yml`, or only in the `baseConfig`. I was specifically trying to avoid that to enforce a certain amount of simplicity in the simple config. Maybe it's not so bad if all of the decision task gunk is elsewhere already though...

rfcs/0036-taskgraph-workflow.md
bhearsum: Supporting this type of organization would be nice, yeah. See the thread above with Dustin for some potential complications :stuck_out_tongue:

rfcs/0036-taskgraph-workflow.md
ahal: Oh I realize now I didn't quite understand Dustin's proposal and mine's very similar.

Except Dustin's proposal would keep backwards compatibility with the existing `.taskcluster.yml` format and namespace all this new stuff we're talking about under an `extra` key or whatever.

In contrast my proposal is talking about breaking backwards compatibility, and instead namespacing all the complex stuff that was previously in `.taskcluster.yml` and having the simple config things be at the top-level.

I still prefer the latter since I think the "simple config" path is the one that should be prioritized, though maybe breaking backwards compatibility here is more trouble than I'm thinking it is.

We both agree that there should be a single config file that contains everything. So as a concrete review comment, I think this RFC should assume the config is still in `.taskclsuter.yml` for now. If we want to change it to `.taskcluster/config.yml` we can, but that can be a separate PR and likely doesn't need an RFC.

rfcs/0036-taskgraph-workflow.md
ahal: Makes sense.. I definitely don't think this needs to be part of the MVP or anything. Let's not support this initially and we can look into it as follow-up work if there is a need.

rfcs/0036-taskgraph-workflow.md
ahal: > My other question is whether or not we'd still support JSON-e in this new version of `.taskcluster.yml`, or only in the `baseConfig`. I was specifically trying to avoid that to enforce a certain amount of simplicity in the simple config. Maybe it's not so bad if all of the decision task gunk is elsewhere already though...

I think only supporting `json-e` for part of the config would be a bad idea. Might as well support it for the entire thing. In the simple case, users could choose not to use it, and our documentation should only mention that its possible under some more advanced sections.

djmitche: We have a `version` property, so we could break compatibility without *too* much trouble.  However, taskcluster-github should still make sense without taskgraph, even in that new version.  I think a decent litmus test there might be to think of what the docs for that functionality might look like, without referencing "taskgraph".

That said, I'm not yet convinced that dropping compatibility is a big win here, and it would be easier to just extend the existing functionality compatibly!

ahal: Oh I didn't consider the need to support `taskcluster-gh` without taskgraph. In that case, yeah I think your approach makes more sense.

In a far off world when we have a solid core taskgraph module that supports simple and complex use cases, I could imagine us unsupporting tc-gh without taskgraph. If that world ever comes to pass, we can always break compatibility to prioritize the simple use case at that point.

imbstack:
This makes a lot of sense to me at a first/second glance. To be completely transparent I don't really have the details of tc-gh paged into my head enough to check the finer points but it all feels doable. I'm happy to help with this however, especially once I have tc-gh time again.

rfcs/0036-taskgraph-workflow.md
escapewindow: URL pros:
- flexibility
- allows for self-serve: if the set of supported templates/baseConfigs don't include what you want, publish your own, and voila.
- allows for pointing at historical revisions/branches/forks

URL downsides:
- encourages lack of standardization
- encourages lack of pinning (point at a branch instead of a revision, and suddenly the reproducibility of a task graph is dependent on the state of an external repository and the timing of when we generated the graph).
    - This may be similar enough to a Dockerfile's `FROM python:3.8` that we're ok with it, but we should consider whether we want more strict pinning.

I'm not sure I 100% agree with what I'm going to propose, but an alternative to URLs may be:

- an overrideable URL to your template repo, with a strong suggestion of using the official template repo. Similar to official deb repos, but with the possibility of pointing at an experimental or user-controlled repo.
- named-and-versioned templates in the template repo(s); the official template repo has a policy of never altering published versions of templates without a version bump

Alternately, use URLs, but:
- mitigate lack of pinning/reproducibility by leaving a breadcrumb in the artifacts that give us specific information about the template we used, up to and including publishing the template we used as a decision task artifact.
- or disallow/discourage the use of unpinned or arbitrary URLs for any level 3 repository

escapewindow: As hand-wavy examples to the above:

```baseConfig: gradle:1.0  # or tcymlTemplate: mobile:1.0, or whatever```

Take the latest templateRepo's revision, find the gradle or mobile template(s), and use the one that matches `1.0`. I haven't yet decided if this means `1.0.1` would work, or `1.x`, or if it explicitly means version `1.0`. If it's a fuzzy version, we at least publish the templateRepo + revision + template path in our decision task artifacts for post-task reproducibility.

Or

```
tcymlTemplateRepo: https://github.com/org/repo
tcymlTemplate: myTemplate:1.0
```

which allows you to maintain your own set. Reproducibility and semver adherence is entirely up to the owners of that repo, which probably means we would discourage or desupport that behavior for level 3 release tasks.

File: rfcs/0036-taskgraph-workflow.md [hunk 7]
```diff
@@ -0,0 +1,258 @@
...
+        produces:
+            macos-unsigned.dmg: build/macos-unsigned.dmg
+            tests.tar.bz2: build/tests.tar.bz2
+    test-macos:
+        workerType: t-macos
+        consumes:
+            # This is a reference to the task above
+            build-macos:
+                - tests.tar.bz2
+            # This is a reference to the task in `kinds/sign/kind.yml`
```
escapewindow, line 192: Hm. Once we come to a soft consensus, it may be worth doing a proof of concept.
I'm a little worried about what this means about kind generation order, but maybe it's solved by an implicit `kind-dependency`. Not sure.

rfcs/0036-taskgraph-workflow.md
djmitche: To sprinkle a few thoughts in here:
 * We currently do not store any per-repo data in tc-github.  In fact, the service has no "write" APIs whatsoever.  We could change that, and add per-repo configuration for how baseConfig is interpreted.  The risk of doing so is that it becomes one more out-of-sight place that things can change.
 * We could also add deployment configuration options for tc-github, if necessary.
 * Supporting semver ranges is probably incompatible with pinning.  You mentioned docker-hub before, and that's a model TC users will be familiar with, so it's probably best to follow its pattern where possible.  And it only supports literal tags, not semver.
 * Docker has a kind of extensible image format that ultimately privileges `hub.docker.com` as the default.  We could do the same, but privilege `github.com` as the default -- this is already the "taskcluster-github" service, so it doesn't seem a stretch.  Then template specs might be
   * `rust` -> `https://github.com/taskcluster-templates/rust/blob/main/taskcluster.yml` (where `main` is the default branch)
   * `rust:1.1` -> `https://github.com/taskcluster-templates/rust/blob/1.1/taskcluster.yml` (here the `taskcluster-templates` org might be a deployment configuration option)
   * `mozilla-releng/mobile:2.3` -> `https://github.com/mozilla-releng/mobile/blob/2.3/taskcluster.yml`
   * `https://myserver.com/mytemplate.yml` -> `https://myserver.com/mytemplate.yml`

It's not clear to me whether `baseConfig` will be interpreted by a user-defined task, or by tc-github.  If the latter, then there is nowhere to automatically attach an artifact for [EDIT 2].  But tc-github could be configured to pass some information to the template via context.  If it's the former, then that task could publish whatever artifacts it wants, including maliciously publishing the correct breadcrumbs.

rfcs/0036-taskgraph-workflow.md
bhearsum: >  I'm a little worried about what this means about kind generation order, but maybe it's solved by an implicit kind-dependency. Not sure.

If nothing else, we can solve this by doing an extra pass (first pass to find all of the names, second pass to do the work).

I think it's important that we allow simple config to depend on non-simple config tasks (otherwise you may end up being forced to rewrite simple tasks when you add a complex one). If we end up in a situation where the use case of "I want to run tests on signed builds" ends up with someone needing to learn full blown Taskcluster - I think that's a failure.

You're rightly identifying an issue when we get to cases of fanning out, or possibly even when we have per-platform tasks in a kind.

I could see requiring simple tasks to specify full task names for non-simple dependencies, and probably supporting a glob to handle cases of locale/chunk fanning out (where using a non-globbed name would be unstable).

This clearly requires a bit more thought though. There's a few concrete cases that we need to handle:
* A simple task depending on _one_ Task generated from a kind (maintaining a fan out that a kind started)
* A simple task depending on all Tasks generated from a kind (fanning in)
* A simple task depending on all Tasks from one job in a kind (eg: fanning in for one platform)
* A simple task depending on all Tasks from multiple jobs in a kind

Are there other cases too?

rfcs/0036-taskgraph-workflow.md
bhearsum: > encourages lack of pinning

In an earlier conversation (I don't recall if you were part of it, @escapewindow), I think we came to the conclusion that full hard pinning (ie: to a revision) is probably not a great idea. I do think we want to version the base configs, but there will be times when we want to fix a bug in them, or add a non-breaking feature (ie: if there's a `tasks_for` to add support for), and I think we want all of the consumers to be able to pick such changes up automatically. This does mean we have to be _very_ careful when making such changes, of course.

> It's not clear to me whether baseConfig will be interpreted by a user-defined task, or by tc-github.

I think it _must_ be intepreted by tc-gh, as the baseConfig is what will define the decision task in most cases. (Maybe I'm misunderstanding what you're saying, though.)

I do like the idea of using Docker-like specs here, both because it may be somewhat familiar, and because it allows the "usual" definitions to be very brief, while providing flexibility to customize. I'd prefer this to having separate entries for repo & template.

rfcs/0036-taskgraph-workflow.md
escapewindow: Your example showed a simple task depending on each of the tasks of _two_ kinds, aiui, where one was a simple kind (a single task) and one was a complex kind (potentially many tasks).

However, I'm fairly certain we can add in one or more common use cases that we can think of, and allow for adding more in the future, either custom ones in `mozilla_taskgraph`, or via submitting a patch upstream if we determine that this use case is generally useful.

rfcs/0036-taskgraph-workflow.md
escapewindow: I'm thinking we should explicitly disallow multiple inheritance of baseConfigs until we have the initial implementation and we can reason about the fallout of such a feature better.

rfcs/0036-taskgraph-workflow.md
ahal: Apologies if this is not the right moment for bike shedding on config names, but I'm not a huge fan of the name `baseConfig` but couldn't think of a better name. But your suggestion of using docker-like specs just sparked the idea of calling this `from` instead, just as you would in a Dockerfile.

escapewindow: > Apologies if this is not the right moment for bike shedding on config names, but I'm not a huge fan of the name `baseConfig` but couldn't think of a better name. But your suggestion of using docker-like specs just sparked the idea of calling this `from` instead, just as you would in a Dockerfile.

I also don't love `baseConfig`, as `base` and `config` are two largely overloaded terms in releng. `from` may not be clear enough? I thought of this as a `template` before this RFC. Maybe we can timebox the bikeshedding :)

escapewindow: > > encourages lack of pinning
> 
> In an earlier conversation (I don't recall if you were part of it, @escapewindow), I think we came to the conclusion that full hard pinning (ie: to a revision) is probably not a great idea. I do think we want to version the base configs, but there will be times when we want to fix a bug in them, or add a non-breaking feature (ie: if there's a `tasks_for` to add support for), and I think we want all of the consumers to be able to pick such changes up automatically. This does mean we have to be _very_ careful when making such changes, of course.

To be clear, if CoT verification can't determine which specific tc.yml we built the decision task from and it guesses wrong, it won't be able to recreate the decision task, and CoT verification will fail. If we don't care about any scriptworker tasks, then this may not matter, but aiui we want to allow for the simple config to work in level 3 repos as well.

We either need to hard-pin, or we need a verifiable breadcrumb (an artifact or decision task `task.extra` breadcrumb) that lets us verify the tc.yml expands to the decision task, and the tc.yml is from a revision in a trusted tree/branch.

> > It's not clear to me whether baseConfig will be interpreted by a user-defined task, or by tc-github.
> 
> I think it _must_ be intepreted by tc-gh, as the baseConfig is what will define the decision task in most cases. (Maybe I'm misunderstanding what you're saying, though.)

I think it's a question of whether tc-gh is doing the logic of finding the baseConfig and creating the decision task, or we call, say, `build-decision` with a hardcoded tc.yml, and that decision-task-creation-task finds the baseConfig and creates the decision task. The latter allows for more customization and longer runtimes (read: complexity), but it means we need to trigger a task that triggers a decision task that creates a graph. We're already doing that for cron and hg-push decision tasks, so expanding build-decision to also handle simple config tasks would just extend that model.

> I do like the idea of using Docker-like specs here, both because it may be somewhat familiar, and because it allows the "usual" definitions to be very brief, while providing flexibility to customize. I'd prefer this to having separate entries for repo & template.

I think I do too, other than knowing that if you change the repo, then you're no longer in our official trust model.

djmitche: Ah, I had forgotten that "breadcrumb" is just a pointer to the thing that should satisfy the validation, and is not itself validated.

escapewindow: >     * We currently do not store any per-repo data in tc-github.  In fact, the service has no "write" APIs whatsoever.  We could change that, and add per-repo configuration for how baseConfig is interpreted.  The risk of doing so is that it becomes one more out-of-sight place that things can change.

I was thinking we could have a hint about how to do this in .tc.yml or .taskcluster/config or whatever, although this means we may need to fetch one or more files from repo, parse, fetch a baseConfig, parse, and then create a decision task. I am starting to lean harder towards "tc-github knows how to launch build-decision, and build-decision knows how to expand simple configs to create an appropriate decision task".

>     * We could also add deployment configuration options for tc-github, if necessary.
> 
>     * Supporting semver ranges is probably incompatible with pinning.  You mentioned docker-hub before, and that's a model TC users will be familiar with, so it's probably best to follow its pattern where possible.  And it only supports literal tags, not semver.

Maybe.
Python deps allow for semver ranges, e.g.

```
Strong equality 	Django==3.0.3
Greater or equal	Django>=3.0.3
Compatible version	Django~=3.0.3
```

However, I understand perfectly if we don't want to hold Python up as a model for handling dependencies :)

What I don't like about docker is the tags can move, so they don't actually hard pin either. If we have a branch or nonofficial url, a moving tag, or a rough range, we still need breadcrumbs to verify the tasks.

> 
>     * Docker has a kind of extensible image format that ultimately privileges `hub.docker.com` as the default.  We could do the same, but privilege `github.com` as the default -- this is already the "taskcluster-github" service, so it doesn't seem a stretch.  Then template specs might be
>       
>       * `rust` -> `https://github.com/taskcluster-templates/rust/blob/main/taskcluster.yml` (where `main` is the default branch)
>       * `rust:1.1` -> `https://github.com/taskcluster-templates/rust/blob/1.1/taskcluster.yml` (here the `taskcluster-templates` org might be a deployment configuration option)
>       * `mozilla-releng/mobile:2.3` -> `https://github.com/mozilla-releng/mobile/blob/2.3/taskcluster.yml`
>       * `https://myserver.com/mytemplate.yml` -> `https://myserver.com/mytemplate.yml`

The more repos we have, the harder it will be to guarantee that we aren't force-pushing or moving tags or similar. That means if we want to hard-pin, we may need to allow for pinning by hash, or maybe we just do the breadcrumb thing, or just have no guarantee of task reproducibility.

> It's not clear to me whether `baseConfig` will be interpreted by a user-defined task, or by tc-github. If the latter, then there is nowhere to automatically attach an artifact for [EDIT 2]. But tc-github could be configured to pass some information to the template via context. If it's the former, then that task could publish whatever artifacts it wants, including maliciously publishing the correct breadcrumbs.

CoT will try to follow the breadcrumbs to get the tc yml template, verify it's in a trusted repo/branch, and recreate the decision task. If the malicious user is able to make all of those things match and still do something malicious, then it will be undetected, but they won't be able to use an arbitrary tc.yml or arbitrary decision task definition.

bhearsum: > To be clear, if CoT verification can't determine which specific tc.yml we built the decision task from and it guesses wrong, it won't be able to recreate the decision task, and CoT verification will fail. If we don't care about any scriptworker tasks, then this may not matter, but aiui we want to allow for the simple config to work in level 3 repos as well.

OK, this is a very good point! CoT is a must-have, so we clearly have to solve this in one of the ways you suggest).

> "tc-github knows how to launch build-decision, and build-decision knows how to expand simple configs to create an appropriate decision task"

If I'm understanding this correctly, this means that `.taskcluster.yml` is not not only handled by tc-gh, but _also_ by a particular decision task we happen to own and implement. I'm not really seeing the benefits of this over tc-gh doing all the parsing, and creating whatever tasks happen to result from it (usually a decision task -- but it doesn't have to be).

I'm starting to think we should hop on zoom to chat about this a bit more too, these's a few things here that are spiraling.

rfcs/0036-taskgraph-workflow.md
bhearsum: Agreed, 100%. I don't think even think we should allow it until we have a specific, known reason that can't be reasonably handled without it.

rfcs/0036-taskgraph-workflow.md
bhearsum: > Your example showed a simple task depending on each of the tasks of two kinds, aiui, where one was a simple kind (a single task) and one was a complex kind (potentially many tasks).

Yeah, I didn't include something like that in the list above because I don't think the simple tasks add additional complexity. You add simple task -> simple task dependencies into any scenario above, but the complexity comes from the simple <-> complex task dependencies -- so I didn't think it was worth adding those variants in (I'm pretty certain they'll end up implemented the same in all the scenarios above).

escapewindow:
Relevant: https://docs.mozilla-releng.net/en/latest/taskcluster/how_tasks_are_triggered.html#side-note-taskcluster-yml-repo

escapewindow:
> From a quick scan of that doc, the approach of allowing arbitrary repos to be referenced is very close to how the GitHub Action ecosystem is implemented. Please ensure that:
> 
>     * pinning is encouraged
> 
>     * versions used are logged somewhere
> 
> 
> I believe in the 1-2 year time frame, we will be wanting to generate SBoMs for both the end artifacts, and the tooling used to create that artifact. So anything being expanded now should be aware, and hopefully not make technical choices that will make SBoM generation impractical.

I believe this approach is limited to level 1 repos, so any pinning etc. would be towards making things more maintainable, as opposed to securing release pipelines.

rfcs/0036-taskgraph-workflow.md
ahal, line 231: Oh, I didn't realize you proposed this here when I filed:
https://github.com/mozilla-releng/releng-rfcs/issues/40

This RFC feels very large and all encompassing.. It might be worth splitting it up into smaller RFCs. E.g, we could have one just for the creation of `mozilla_taskgraph`. Though tbh, creation of `mozilla_taskgraph` is somewhat tangential to this RFC. I think doing that first will be less work in the end, but it's not strictly necessary.

bhearsum:
This has been been partly replaced with other ideas, and some parts of it we've decided to abandon for now.
https://bugzilla.mozilla.org/show_bug.cgi?id=1815284, https://github.com/taskcluster/taskgraph/issues/184, and https://github.com/taskcluster/taskgraph/issues/186 are all smaller ways in which we may be able to simplify task configuration. There's also an idea floating around to eliminate the need for projects to provide their own `.taskcluster.yml` (which will likely eventually end up as a Taskcluster RFC).

PR #39, opened 2022-02-09
RFC 37: Installing Decision task Python dependencies from .taskcluste…

ahal:
…r.yml

### DIFF
--- a/rfcs/0037-decision-python-deps.md
+++ b/rfcs/0037-decision-python-deps.md
@@ -45,5 +45,10 @@ While this won't be enforced, the recommended location for this file will be
    point at our internal pypi instead. It's worth noting that any repo which installs taskgraph via
    `pip install path/to/taskgraph` (which appears to be all of them) already depend on pypi.
 
+2. We'll need to be cautious not to break hooks, since previously the version of taskgraph being
+   used was baked into the `.taskcluster.yml`. Since changes to this file invalidate hooks, an
+   update to taskgraph will no longer force us to re-generate them. I don't believe this will cause
+   any problems, but extra testing should be conducted here.
+
 
 [0]: https://pypi.org/project/taskcluster-taskgraph/
--- a/rfcs/0037-decision-python-deps.md
+++ b/rfcs/0037-decision-python-deps.md
@@ -39,6 +39,23 @@ E.g:
 While this won't be enforced, the recommended location for this file will be
 `taskcluster/requirements.txt`.
 
+## Testing Taskgraph Changes
+
+Previously testing changes to taskgraph within a consumer repo was easy. Just edit the repo to point
+at `taskgraph-try` and select whichever revision is necessary. This proposal will complicated the
+process a little bit.
+
+The new recommended way to test changes is to use `pip`'s built-in [VCS support][1]. For example,
+the following specifier could be used to test arbitrary taskgraph revisions:
+
+```
+taskcluster-taskgraph@hg+https://hg.mozilla.org/ci/taskgraph@ae7697ec7905216c7245bafb8a9475355ea00a76
+```
+
+This can be compiled with hashes by `pip-compile` or used as a specifier to `hashin`. The new
+process for testing against arbitrary taskgraph revisions will be documented in taskgraph's
+documentation.
+
 # Open Questions
 
 1. If a repository does not wish to have a dependency on pypi, we may need to provide a method to
@@ -52,3 +69,4 @@ While this won't be enforced, the recommended location for this file will be
 
 
 [0]: https://pypi.org/project/taskcluster-taskgraph/
+[1]: https://pip.pypa.io/en/latest/topics/vcs-support/
--- a/rfcs/0037-decision-python-deps.md
+++ b/rfcs/0037-decision-python-deps.md
@@ -6,11 +6,11 @@ Outline a standard best practice for installing Python dependencies from a `.tas
 
 ## Motivation
 
-Many Taskcluster enabled repositories depend on Python via their use of taskgraph. Taskgraph is
+Many Taskcluster enabled repositories depend on Python via their use of Taskgraph. Taskgraph is
 typically bootstrapped via the `run-task` script which clones the repository and checks out a
 specified revision. This isn't ideal because it's not easy to tell how out of date the version of
-taskgraph being used is. It's also difficult to determine if there are any breaking changes while
-updating. Both of these problems would be solved by using the [pypi releases][0] which follow
+Taskgraph being used is. It's also difficult to determine if there are any breaking changes while
+updating. Both of these problems would be solved by using the [Pypi releases][0] which follow
 semantic versioning.
 
 Some repositories also depend on other libraries like `mozilla_version` or `redo`. These are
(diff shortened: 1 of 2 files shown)


rfcs/0037-decision-python-deps.md
```diff
@@ -0,0 +1,54 @@
...
+# Open Questions
+
+1. If a repository does not wish to have a dependency on pypi, we may need to provide a method to
+   point at our internal pypi instead. It's worth noting that any repo which installs taskgraph via
+   `pip install path/to/taskgraph` (which appears to be all of them) already depend on pypi.
+
+2. We'll need to be cautious not to break hooks, since previously the version of taskgraph being
+   used was baked into the `.taskcluster.yml`. Since changes to this file invalidate hooks, an
+   update to taskgraph will no longer force us to re-generate them. I don't believe this will cause
+   any problems, but extra testing should be conducted here.
```
escapewindow, line 51: 3. We currently point at a revision to, say, test against [taskgraph-try](https://hg.mozilla.org/ci/taskgraph-try/). How will we support this option alongside the pip install? (e.g. the main repo uses a pip install with pinned requirements.txt; we want to test something; does testing against taskgraph-try involve significant .tc.yml hacking?)

tomprince: It is possible to refer to a specific URL in a `requirements.txt` file, with something like
```
taskcluster-taskgraph @ https://hg.mozilla.org/ci/taskgraph-try/archive/41201f4ed1c23bec76cc24e41b3f46d73e941a94.zip \
    --hash=sha256:ec8675522b82e9487c42a0ef3417682fa3a0a127ca7fa05678df233c983ab9f3
```
so it shouldn't involve any `.taskcluster.yml` hacking.

ahal: Good point, this is worth calling out in the RFC. I think users could edit the `requirements.txt` to use a link to the repo instead:
https://pip.pypa.io/en/latest/topics/vcs-support/#mercurial

Revision numbers, tags and branches are all supported as well. So the only difference is instead of updating the revision in `.taskcluster.yml` a user would update the specifier in `requirements.in` (or whatever is used to generate the `requirements.txt`).

I verified and the following requirements.txt is valid:
```
taskcluster-taskgraph@hg+https://hg.mozilla.org/ci/taskgraph@ae7697ec7905216c7245bafb8a9475355ea00a76
```
At least `hashin` can also generate hashes for this specifier as well.

ahal: This is less intuitive though, so good documentation will be needed..

ahal: (Oh, I mid-aired with Tom but we said the same thing :), thanks Tom! And good to see you)

rfcs/0037-decision-python-deps.md
```diff
@@ -0,0 +1,54 @@
...
+(as long as it has hashes).
+E.g:
+
+1. Generate it manually with help from `hashin`.
+2. Compile it with `pip-tools`.
+3. Export it from `poetry`.
+4. Etc.
+
+While this won't be enforced, the recommended location for this file will be
+`taskcluster/requirements.txt`.
```
tomprince, line 40: Would it make sense to looks for this file by default?

ahal: Yeah, I thought about that, and it might? I think the tradeoff here is that it might be a bit too magical. Someone trying to figure out how `taskgraph` gets installed will need to know to look in `run-task`, whereas needing an env leaves a breadcrumb for them to follow. I could be persuaded to go either direction.

rfcs/0037-decision-python-deps.md
```diff
@@ -0,0 +1,54 @@
...
+
+# Open Questions
+
+1. If a repository does not wish to have a dependency on pypi, we may need to provide a method to
+   point at our internal pypi instead. It's worth noting that any repo which installs taskgraph via
+   `pip install path/to/taskgraph` (which appears to be all of them) already depend on pypi.
```
tomprince, line 46: If `run-task` commits to using pip to install the requirements, then consumer could set `PIP_INDEX_URL` in the environment to point at a different index, or `PIP_NO_INDEX=1` to disable the default index (plus `PIP_FIND_LINKS` to point at presumably https://pypi.pub.build.mozilla.org/pub/).

rfcs/0037-decision-python-deps.md
ahal: Another benefit of the env, is this feature might be useful outside of decision tasks. E.g, it could be used to install test dependencies in a test task for instance. There are plenty of other options there though, so not super compelling.. maybe there's more compelling use cases I'm not thinking of.

rfcs/0037-decision-python-deps.md
escapewindow: +1, if this works with a hash-pinned requirements.txt, and if we document it, wfm :)

rfcs/0037-decision-python-deps.md
```diff
@@ -0,0 +1,54 @@
...
+Some repositories also depend on other libraries like `mozilla_version` or `redo`. These are
+typically installed via `pip install` on the Decision task's command line. This isn't ideal because
+we won't verify hashes and the version is often unpinned.
+
+We should have a consistent method of installing Python dependencies which covers both use cases and
+addresses the downsides of the current solutions for each.
+
+# Details
+
+The `run-task` script will be updated to support a `<REPO>_PIP_REQUIREMENTS` environment variable.
```
tomprince, line 25: I'd be inclined to document how having multiple repos with the corresponding environment variable will be handled. Also, this should specify at what point this occurs (relative to e.g. fetches; if it is is after, then requirements files could include `$MOZ_FETCHES_DIR` in them to refer to fetches).

rfcs/0037-decision-python-deps.md
```diff
@@ -0,0 +1,54 @@
...
+This will be a relative path from the root of the repository specified by `<REPO>` pointing to a
+`requirements.txt` file. This `requirements.txt` file will be installed into the active Python
+environment via `python -mpip install --require-hashes -r $<REPO>_PIP_REQUIREMENTS`.
+
+Taskcluster consumers are free to use whichever tools they see fit to manage the `requirements.txt`
+(as long as it has hashes).
```
tomprince, line 31: Even if all of these work, would it make sense to specifying a suggested way of managing it, even if there are multiple options? (I guess this would probably be pip-tools).

rfcs/0037-decision-python-deps.md
```diff
@@ -0,0 +1,54 @@
...
+
+We should have a consistent method of installing Python dependencies which covers both use cases and
+addresses the downsides of the current solutions for each.
+
+# Details
+
+The `run-task` script will be updated to support a `<REPO>_PIP_REQUIREMENTS` environment variable.
+This will be a relative path from the root of the repository specified by `<REPO>` pointing to a
+`requirements.txt` file. This `requirements.txt` file will be installed into the active Python
+environment via `python -mpip install --require-hashes -r $<REPO>_PIP_REQUIREMENTS`.
```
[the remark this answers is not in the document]
ahal: True, I agree this is an area that could use some work. Though I'd like to leave it out of this RFC for now as this RFC shouldn't impact anything. Previously we just installed things into whatever environment the `pip3` binary was part of, and we'll continue doing the same here (just using `python3 -mpip` instead).

rfcs/0037-decision-python-deps.md
ahal: Yeah I think that makes sense, at least in the docs (should it also be in the RFC?). I agree we'll likely use `pip-compile` for anything releng-owned.

rfcs/0037-decision-python-deps.md
escapewindow: This wfm, thank you!

escapewindow (approved):

PR #41, opened 2022-11-02
RFC 41: Create a repo for Mozilla-specific Taskgraph logic

ahal:
See the [rendered proposal here](https://github.com/ahal/releng-rfcs/blob/mozilla_taskgraph/rfcs/0041-mozilla-taskgraph.md)!

### DIFF
--- a/rfcs/0041-mozilla-taskgraph.md
+++ b/rfcs/0041-mozilla-taskgraph.md
@@ -1,0 +1,136 @@
+# RFC 0041 - Mozilla Taskgraph
+
+* Comments: [#40](https://github.com/mozilla-releng/releng-rfcs/issues/40)
+* Proposed by: @ahal
+
+## Summary
+
+A new Python package called `mozilla-taskgraph` will be created to contain
+Mozilla-specific Taskgraph logic currently residing in the [upstream Taskgraph
+module][taskgraph] and various project Taskgraphs (especially in mobile).
+
+### Motivation
+
+This change will be beneficial both from the perspective of Mozilla as a
+Taskgraph consumer, as well as Taskcluster as a standalone open-source project.
+
+#### Motivation for Mozilla
+
+Taskgraph is used across a variety of projects, each with unique needs and
+unique tasks to support those needs. But there are also a lot of commonalities
+between these projects. If Taskgraph logic should be shared across
+projects, the only current place to do so is in the upstream [Taskgraph]
+repository. Trying to add Mozilla-specific logic to this package, while at the
+same time not breaking consumers outside of Mozilla is a big maintenance burden.
+
+Having a dedicated place to add Mozilla logic will make it easier Releng and
+devs using Taskgraph to organize code without worrying about impacting external
+consumers.
+
+#### Motivation for Taskcluster
+
+Taskgraph is a unique tool and very good at generating a large graph of tasks.
+In most ways it is the ideal complement for Decision tasks to call. The main
+caveat is that it is currently only suitable to use at Mozilla, as there is a
+lot of Mozilla-specific logic and terminology baked-in.
+
+By pulling out the Mozilla-specific logic from core Taskgraph, Taskcluster can
+finally point to Taskgraph as the recommended / standard way for consumers to
+implement a Decision task.
+
+## Details
+
+### Project
+
+The source for the new `mozilla-taskgraph` package will live under
+`https://github.com/mozilla-releng/mozilla-taskgraph` and be published to Pypi
+under the same name.
+
+The `mozilla-taskgraph` project will use the same test, lint and code coverage
+infrastructure as `taskcluster-taskgraph`.
+
+### Dependency Management
+
+The `mozilla-taskgraph` package will use the same mechanism as core Taskgraph
+to manage its dependencies. As of this writing, this is via
+`pip-compile-multi`.
+
+The `mozilla-taskgraph` package will depend on the `taskcluster-taskgraph`
+package and pin it using the [compatible release operator]. This will ensure
+that consumers of `mozilla-taskgraph` can't accidentally use a version of
+`taskcluster-taskgraph` that is incompatible with it.
+
+The `mozilla-taskgraph` maintainers will need to create a new version whenever
+a major upgrade of `taskcluster-taskgraph` is released.
+
+### Source Layout
+
+The `mozilla-taskgraph` package will have a layout similar to:
+
+```bash
+mozilla_taskgraph
+├── morph.py
+├── optimize
+│   └── strategies.py
+├── parameters.py
+├── target_tasks.py
+└── transforms
+    ├── mobile
+    │   ├── some_transform.py
+    │   └── util
+    └── scriptworker
+    │   ├── some_transform.py
+    │   └── util
+    └── util
+
+```
+
+This is similar to core Taskgraph with a few differences:
+
+1. Each "domain" will have a dedicated directory under the top-level
+   `transforms` directory.
+2. Utility files (if any) will live alongside the transforms that use them. If
+   needed, a global utility module will exist for sharing across domains.
+
+### Consumer Integration
+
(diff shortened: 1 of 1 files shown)


ahal:
[Rendered](https://github.com/ahal/releng-rfcs/blob/mozilla_taskgraph/rfcs/0041-mozilla-taskgraph.md)

rfcs/0041-mozilla-taskgraph.md
```diff
@@ -0,0 +1,130 @@
...
+
+```bash
+mozilla_taskgraph
+├── morph.py
+├── optimize
+│   └── strategies.py
+├── parameters.py
+├── target_tasks.py
+└── transforms
+    ├── mobile
```
bhearsum, line 77: What do you think about naming domains after what they do rather than what they're currently used by? With this in mind, I agree that `scriptworker` is very sensible, but I'm not sure about `mobile`. If we, eg, ever start doing non-mobile things with gradle, this could get confusing. `gradle` would certainly be a sensible one though. 

This might be worrying too much about something that's not a problem yet though - so feel free to ignore.

rfcs/0041-mozilla-taskgraph.md
```diff
@@ -0,0 +1,130 @@
...
+## What is Mozilla-specific?
+
+What is and isn't Mozilla-specific could be an open debate. The following list
+is *very roughly* ordered from least to most controversial, and only contains
+items that already exist in Taskgraph core:
+
+- Anything related to scriptworkers
+- Anything related to Treeherder
+- Code review tasks
+- The concept of `level`
```
bhearsum, line 112: Oh boy, everything here and down I could make a very strong argument for being in taskgraph! I don't think we ought to block on it (we can always move things back to `taskgraph` later if we want) - but I'm not sure we yet know how opinionated we want `taskgraph` to be -- and perhaps that's just as much of a question for the Taskcluster team as it is for us.

bhearsum (commented):
This is very well written and covers pretty much everything I can think of. I have a couple of small things inline, but the one thing I'm not clear on after reading this is whether or not `taskgraph` will remain the entry point to decision task (and `mozilla_taskgraph` is something that can be fed to it somehow), or whether `mozilla_taskgraph` will be the entry point to decision tasks (and it will depend on `taskgraph`. If it's the former, an explanation of how that will work is probably warranted. If it's the latter, it ought to be made explicit.

rfcs/0041-mozilla-taskgraph.md
JohanLorenzo: I agree. On the one hand, I feel like standalone Taskgraph should be opinionated about how tasks should be split. Based on the experience we had in Android - when we had our custom implementation of a decision task, we ended up needing much of these concepts. I wouldn't be surprised if external consumers would go down the same route. 

On the other hand, if taskgraph alone is too complex then I'm not sure how many external consumers would be ready to make the jump.

I guess we have to strike a balance between "taskgraph is easy enough to use" and "taskgraph is powerful enough to support the hardest use case". How about we put these listed item to `mozilla_taskgraph` first. Then, whenever the first external consumer comes in and want a feature not in standalone taskgraph anymore, we just port it back? This way, we can help create a sense of ownership from our first consumers (i.e.: "I made this feature happen so I will use it")

rfcs/0041-mozilla-taskgraph.md
JohanLorenzo: I'm fine naming some folders by their trust-domain. I'm thinking of the `build_apk` transform for instance. It's definitely `mobile` and I hardly see a case where we use it outside of a project that builds some Android apps. 

That said, for sure we can use `gradle` for something that is unrelated to Android. As far as I know we don't have any known case at Mozilla. I guess we can cross that bridge when we come to it.

JohanLorenzo (approved):
Looks great to me! I think we're lucky to be in a position where we are currently the sole consumers. This means we can still port code back and forth between `taskgraph` and `mozilla_taskgraph`, depending on the most recent learning.

Thus, I'm okay to get started with this proposal without specifying too much 👍

rfcs/0041-mozilla-taskgraph.md
ahal: Yeah, I agree with breaking them down by function rather than project area. I also agree with Johan that there's likely enough mobile-specific stuff that `mobile` could be a trust domain *or* a function. Some alternative ideas to `mobile` might include `android` / `ios`, or `build.android` / `build.ios` / `test.android` / `test.ios`, etc.

But the layout here was just an example, and not necessarily what we'll end up with. I think we can discuss the finer points in the PRs that actually migrate the logic over.

rfcs/0041-mozilla-taskgraph.md
ahal: Yeah, there's definitely a lot open to debate! I think to start I'd like to focus on just the obvious stuff, and then we can refactor going on from there.

I think one strategy would be to try and genericize these features. For example, we could rename `level` to `trust_level`, and then allow configuring how many trust levels there are for a given project (setting this to 1 would effectively turn the feature off). You could imagine a simple Github project might only have two trust levels (PR + push), whereas a more complex one might have 4 (untrusted PR, PR, push, release).

For `trust_domain`, if unspecified we could use a generically named pools / scopes / etc, but if set add the `trust_domain` in.

ahal:
> 



> This is very well written and covers pretty much everything I can think of. I have a couple of small things inline, but the one thing I'm not clear on after reading this is whether or not `taskgraph` will remain the entry point to decision task (and `mozilla_taskgraph` is something that can be fed to it somehow), or whether `mozilla_taskgraph` will be the entry point to decision tasks (and it will depend on `taskgraph`. If it's the former, an explanation of how that will work is probably warranted. If it's the latter, it ought to be made explicit.

Good call out! I was thinking it would be the former. This would be consistent with how the various `project_taskgraph` modules work. The main entrypoint is still upstream Taskgraph, but they each have a `register()` function which sets up things like custom parameters / morphs / etc. Transforms are easier because you just have to import them.

That said, the [Gecko __init__.py](https://searchfox.org/mozilla-central/source/taskcluster/gecko_taskgraph/__init__.py#1) has a lot of hacky monkeypatching needed to get things working. So I think a part of this work will involve improving Taskgraph's entry points for consumers to hook into so stuff like that isn't needed anymore.

rfcs/0041-mozilla-taskgraph.md
bhearsum: Oh sorry, I didn't mean to bikeshed unnecessarily!

rfcs/0041-mozilla-taskgraph.md
bhearsum: This sounds like a very reasonable approach -- both the idea of starting with the obvious things, and trying to make things more generic. Better to be conservative about what goes in though -- it's much harder to remove or change things later, if, eg: it's being depended on by a bunch of people.

bhearsum (approved):

jfx2006:
This is great! Thunderbird abuses taskgraph in interesting ways. This seems like it would really help.

rfcs/0041-mozilla-taskgraph.md
```diff
@@ -0,0 +1,139 @@
...
+
+* Anything related to scriptworkers
+* Anything related to Treeherder
+* Code review tasks
+* The concept of `level`
+* The concept of `trust_domains`
+* The concept of `project`
+* Index route format (in `transforms/task.py`)
+* Toolchains
+* ...
```
lotas, line 128: generated `actions.json` highly depends on hooks with specific names, which are deployed/managed independently. 
Would be good to make that part also moz-specific

lotas: I'm not saying we should remove this feature, just maybe make it more customizable or more generic. Feature is useful, agree.
Thanks!

PR #44, opened 2023-09-07
Regular deploys of Balrog & Ship It

bhearsum:
[rendered](https://github.com/mozilla-releng/releng-rfcs/blob/5ddb8281d1507b3b66ef540c49473fd17a422a7c/rfcs/0043-regular-deploys.md)

### DIFF
--- a/rfcs/0044-regular-deploys.md
+++ b/rfcs/0044-regular-deploys.md
@@ -1,0 +1,40 @@
+# RFC 44 - Scheduled Updates & Deploys of Balrog, Ship It, and Scriptworkers
+* Comments: [#44](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/43)
+* Proposed by: @bhearsum
+
+# Summary
+
+We should regularly update and deploy our most crucial services to product: Balrog, Ship It and Scriptworkers.
+
+## Motivation
+
+As these projects have matured and stabilized the time between deploys has lengthened. Not exercising our deployment pipeline regularly increases the risk that it will be broken at an inopportune time (such as when deploying an urgent fix). Most recently, we had an example of [Balrog's kubernetes configs needing a bump](https://github.com/mozilla-services/cloudops-infra/pull/5041) before deploys could successfully complete.
+
+Additionally, we have not been keeping up on dependency updates in either of these projects. This both increases the risk of security issues cropping up, and the risk of getting far enough out of date that we can no longer take _any_ dependency upgrades without major headaches or overhaul.
+
+# Details
+
+## Backend Dependencies Only
+
+For the time being, we will only updated backend dependencies as part of these. We have much more limited frontend knowledge, and zero tests for frontend - so regular updating of these dependencies poses a much higher risk of bustage, and when it happens, will be more difficult to isolate and fix.
+
+## Process Changes
+
+Few changes will be needed to implement this proposal. We already have tools and processes for updating dependencies and deploying new versions - the main thing needed will be reminders for those on releaseduty. To this end, we will add two new Slack notifications targeting @releng:
+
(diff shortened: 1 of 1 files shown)


gabrielBusta:
Thought - should we also do s-s? We could batch that with shipit.

RE the front-end dependencies on shipit and balrog I could take a stab at it.

bhearsum:
> Thought - should we also do s-s? We could batch that with shipit.

I can definitely see an argument for doing scriptworker-scripts too, yeah! I totally forgot about those when writing this, but they are obviously the other major piece of infra.

> RE the front-end dependencies on shipit and balrog I could take a stab at it.

Can you clarify what you mean here? Even if we were to do a one time update of the dependencies (which is a good idea IMO), doing regular updates after that still carries higher bustage risk due to lack of tests.

gabrielBusta:
> Can you clarify what you mean here? Even if we were to do a one time update of the dependencies (which is a good idea IMO), doing regular updates after that still carries higher bustage risk due to lack of tests.

Oh I see. I was thinking on a tangent about updating React/MUI to the next major version (I see they are still getting security patches so it's not to pressing.) Tests would come in handy :/

bhearsum:
> > Can you clarify what you mean here? Even if we were to do a one time update of the dependencies (which is a good idea IMO), doing regular updates after that still carries higher bustage risk due to lack of tests.
> 
> Oh I see. I was thinking on a tangent about updating React/MUI to the next major version (I see they are still getting security patches so it's not to pressing.) Tests would come in handy :/

I think that's a good thing to do regardless! I'm just not sure doing so would affect anything here (but I could be convinced otherwise).

rfcs/0043-regular-deploys.md
```diff
@@ -0,0 +1,42 @@
...
+Few changes will be needed to implement this proposal. We already have tools and processes for updating dependencies and deploying new versions - the main thing needed will be reminders for those on releaseduty. To this end, we will add two new Slack notifications targeting @releng:
+
+1) A reminder 2 days before the expected deploy date to update the dependencies and deploy to stage. This will give us time to sanity check and ideally fix any issues.
+2) A reminder on deploy day to perform the production deployment.
+
+The other change we will make as part of this is to ensure that RelEng can complete Balrog deployments without the involvement of the SRE team. (This is not strictly necessary to get where we want, but it avoids extra busy work for SREs, and moves in the direction that they would like to go anyways.)
+
+## Schedule
+
+We will do these deploys during the same weeks as Firefox CI Taskcluster deploys, but on different days. Taskcluster deploys happen on Wednesdays, so we should stage these deployments on Tuesdays, and push to production on Thursdays.
```
ahal, line 28: Firefox-CI Taskcluster deploys are going to be moving to Thursdays due to the new beta schedule. Maybe we should just do these the opposite week as fxci deploys?

rfcs/0043-regular-deploys.md
```diff
@@ -0,0 +1,42 @@
...
+
+# Open Questions
+
+* Are we happy with the proposed schedule above?
+* Will we grant RelEng the necessary access in Jenkins to click the deploy button, or remove the need for SREs in some other way?
+* Should we update backend and frontend dependencies? Or just backend? (Frontend upgrades carry more risk because we have no tests for this code.)
```
ahal, line 34: Maybe just backend for now until we are more confident with front-end?

ahal (approved):
I agree with this proposal. Do we want to include scriptworker-scripts? I think more deploys in general isn't necessarily a bad thing.

bhearsum:
> Do we want to include scriptworker-scripts? I think more deploys in general isn't necessarily a bad thing.

Yeah, I think we should. I'll update the proposal to reflect this.

rfcs/0043-regular-deploys.md
bhearsum: Yes - that's a great idea!

PR #50, opened 2025-03-11
RFC 0050 - System Addons release process

ahal:
[Rendered](https://github.com/ahal/releng-rfcs/blob/ahal/push-umprkkprronq/rfcs/0050-system-addons-release-process.md)

### DIFF
--- a/rfcs/0050-system-addons-release-process.md
+++ b/rfcs/0050-system-addons-release-process.md
@@ -1,0 +1,114 @@
+# RFC 0050 - Updated System Addons release process
+* Proposed by: @ahal
+
+# Summary
+
+The system addons release process was designed for a different time. It
+involves people who don't need to be involved, and omits people who should be.
+
+This RFC aims to streamline the process such that we can both reduce friction
+and increase security.
+
+## Motivation
+
+System addons have largely gone unused over the past several years, but
+recently activity is picking up. They are actively being used for webcompat
+interventions, which can sometimes see multiple releases a month.
+
+They are also planned to be used by the new tab team and others.
+
+## Out of Scope
+
+There is an adjacent question around where System Addons live. Currently their
+canonical location is in the `mozilla-extensions` Github org, though some teams
+are landing code in `mozilla-central` and then syncing it over. As annoying as
+this is, this question is considered out of scope for this RFC. I'd like to
+focus solely on the release mechanics and leave this issue as follow-up work.
+
+# Details
+
+## Repository
+
+The process starts with protecting the repository source. We need to guarantee:
+
+- Only authorized users can push changes.
+- That all code changes are properly reviewed.
+
+To this end, all teams working on system addons must create a team in the
+`mozilla-extensions` org. This team should have write access to their add-on
+repo and the following branch protections must be applied at a minimum:
+
+1. Require signed commits
+2. Require a pull request before merging
+   - Dismiss stale pull request approvals when new commits are pushed
+   - Require approval of the most recent reviewable push
+
+3. Block force pushes
+
+## Shipit
+
+### Auth
+
+By default, Releng will have the ability to sign off on all System Addon
+phases. This is so we can ensure the ability to ship them in an emergency.
+
+But for long lived system addons that are actively being maintained, it will be
+required to create an LDAP group *per System Addon*. This group should contain
+members of the team working on the System Addon. Ideally there should be at
+least 3-4 people in this group as dual signoff will be required and this helps
+account for timezones and folks being on PTO. Membership in this group grants
+the ability to create and cancel releases, as well as schedule and sign off on
+all phases of the release. As noted above, Releng will be able to sign off in
+the event there aren't enough developers around, but this should only be done
+in an emergency, and after due diligence has been undertaken to ensure the
+request is legitimate.
+
+For System Addons that are one-off / short lived, an LDAP group is not
+necessary and Releng signoff will be sufficient. In this event, the request
+should come out of an active incident channel.
+
+### Phases
+
+1. Build
+   - Builds the addon
+   - No signoffs required
+
+2. Ship
+   - Release signs the build, pushes to archive.m.o, creates a release in Balrog
+   - Two signoffs from System Addon specific LDAP group required
+
+Note: The ship phase does not actually ship the system addon to users, this is
+handled in Balrog where the release is handed off to Releng.
+
+## Balrog
+
+This RFC recommends no changes to the process within Balrog, but they are listed
+anyway for completeness:
+
+1. Upon request, Releng creates the appropriate superblob and sets of the rule
+   on the desired `<channel>-sysaddons` channel(s).
+2. QA performs update testing to verify everything is working as expected.
+3. Upon request, Releng duplicates the appropriate rule(s) over to the
+   `<channel>` channel(s).
+   - This requires 1 releng signoff (to validate the rules) and 1 relman
+      signoff (to greenlight the timing)
+
+# Open Questions
+
+1. Should Releng have the ability to do Shipit signoffs in the event of an
+   emergency? 
+2. Is "Ship" the right name for the phase in Shipit? It's shipped in the sense
+   that it is release signed and exists on archive.m.o, though not shipped to
+   users.
+3. Should there be two releng signoffs in Balrog? Creating superblobs and
+   shipping them in the desired places can be tricky, so maybe we should
+   prevent the ability of someone signing off on their own changes.
+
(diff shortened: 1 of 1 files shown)


ahal:
This doc outlines the new process, but doesn't really cover what would be changing from the current process. The main changes from the current process are:

1. Stricter enforcement of Github branch protections
2. Merge the Promote + Ship phases into a single phase
3. Create per-addon LDAP groups and require dual signoff within this group at the Ship phase

rfcs/0050-system-addons-release-process.md
```diff
@@ -0,0 +1,103 @@
...
+   on the desired `<channel>-sysaddons` channel(s).
+2. QA performs update testing to verify everything is working as expected.
+3. Upon request, Releng duplicates the appropriate rule(s) over to the
+   `<channel>` channel(s).
+   - This requires 1 releng signoff (to validate the rules) and 1 relman
+      signoff (to greenlight the timing)
+
+# Open Questions
+
+1. Should Releng have the ability to do Shipit signoffs in the event of an
```
gabrielBusta, line 87: I lean to no on this one. Has there ever been a situation where release engineering had to ship a system add-ons in an emergency?

If it did happen, and we need to ship system add-ons without developer sign-offs, the change to shipit is trivial (we just add our admin LDAP group to the sign-offs for the addon I think.) That's considerable friction for us in that situation though.

rfcs/0050-system-addons-release-process.md
```diff
@@ -0,0 +1,103 @@
...
+   emergency? 
+2. Is "Ship" the right name for the phase in Shipit? It's shipped in the sense
```
gabrielBusta, line 89: It is misleading. Ship makes me think "release to end users running Firefox." Maybe it could be called "push" or "promote."

rfcs/0050-system-addons-release-process.md
```diff
@@ -0,0 +1,103 @@
...
+   that it is release signed and exists on archive.m.o, though not shipped to
+   users.
+3. Should there be two releng signoffs in Balrog? Creating superblobs and
```
gabrielBusta, line 92: The couple of time I've shipped these "super blobs" were not required.
I wonder why they were added in the first place. From what I can gather a super blob is a release in Balrog that represents a group of releases. So I guess it might be a way for the client to download a lot of add-ons group together? From what I've done with these it seems one add on build corresponds to one Balrog release. I don't know if there's system add-ons that are a group of add-ons, or if sometimes you'd want to ship different add-ons together for technical reasons. The rules to target the right Firefox versions with the system add-ons are tricky, but the "create a release" bit not so much.

rfcs/0050-system-addons-release-process.md
```diff
@@ -0,0 +1,103 @@
...
+
+Note: The ship phase does not actually ship the system addon to users, this is
+handled in Balrog where the release is handed off to Releng.
+
+## Balrog
+
+This RFC recommends no changes to the process within Balrog, but they are listed
+anyway for completeness:
+
+1. Upon request, Releng creates the appropriate superblob and sets of the rule
```
gabrielBusta, line 77: A little bit off topic. But it would be cool to automate this.

I can imagine:

Balrog script tasks that create a release in Balrog for the system add-on build to be shipped. Then release engineering can create rules for the developers to get the add-ons on the right Firefox versions.

gabrielBusta: In general, the rules for targeting system add-ons releases are ad-hoc and don't seem follow a particular pattern (this is an assumption, and actually come to think of it, I think I recall webcompat's version has some relationship to gecko's version that can probably be used to derive what the rule is supposed to be.)

gabrielBusta: Another thought:

What if, like we declare the versions in the system add-on extension manifest, system add-on developers could fill out what Firefox releases a version is supposed to target? It would probably be on a different file. Maybe a file that keys from add-on release to Firefox targets. Then there could be Balrog script tasks that create the rule, and then a couple of release engineers sign-off on that in Balrog. Usually it is the developers that know which Firefox versions are supposed to get the out-of-band updates. On the other hand, the way Balrog rules work can be counter-intuitive.

rfcs/0050-system-addons-release-process.md
ahal: Yeah, basically they let us ship multiple add-ons to a single Firefox version. Right now there's only really one add-on using this mechanism, but sounds like there will be more in the future. So we might need to add / remove system addons from a non-empty superblobs in the future.

I think the alternative would be to make each system addon a top level product in balrog, but this would mean a lot of changes to balrog and the update client. It would also add friction to new system addons starting out. Since the superblob is already implemented I don't think it's worth revisiting how this works.

rfcs/0050-system-addons-release-process.md
ahal: Yeah, automating this will be tricky because every release is going to be targeting unpredictable channels / versions. So we'd need to specify what channels / versions are expected in the task payload somehow.

We would also still need to change rules manually, e.g to stop shipping a system addon or to ship it in new channels / versions after the fact. There's often a lot of special timing with these things.

Maybe a more beneficial use of time would be improving morgoth and our ability to make adhoc tweaks instead of focusing on full automation.

rfcs/0050-system-addons-release-process.md
```diff
@@ -0,0 +1,103 @@
...
+   - Dismiss stale pull request approvals when new commits are pushed
+   - Require approval of the most recent reviewable push
+
+3. Block force pushes
+
+## Shipit
+
+### Auth
+
+An LDAP group must be created *per System Addon* that contains the members of
```
jcristau, line 52: I wonder if this has to be ldap or if people.m.o groups would also work.  Probably out of scope for this, since shipit currently doesn't know about the latter.

rfcs/0050-system-addons-release-process.md
jcristau: I'd rather it be allowed out of the box than have to make an emergency shipit change and deploy in the event it's required, if history is any indication it'll be required at some point anyway...

rfcs/0050-system-addons-release-process.md
jcristau, line 92: Leaning towards no.  It's currently possible to request additional review if/when we're unsure, and making an extra signoff required increases the bus factor in the event of an urgent hotfix.  I could be convinced otherwise though.

jcristau (approved):
SGTM

rfcs/0050-system-addons-release-process.md
ahal: I don't have a problem with people.m.o, but that might need an RRA as it's relaxing security? But yeah, either way it's out of scope here.

ahal:
Thanks for the comments, moving this to the proposal phase.

rfcs/0050-system-addons-release-process.md
bhearsum: For comparison, "Ship" for Firefox means more or less the same thing (although it schedules a change in Balrog that will eventually make updates live). I'd be fine with a new name, but personally I think "Ship" is fine here.

bhearsum:
Do we need stakeholders outside of RelEng to weigh in here?

gabrielbusta (approved):

ahal:
I sent the link to #system-addon-release-process (new channel Donal created) a couple times. I'll ping a some folks directly near the end of the week if they haven't commented by then.

ahal:
Recent events made me realize that sometimes we ship one-off / short lived system addons, e.g to address a specific incident. We probably want to take a slightly different process for these (e.g no need to create LDAP groups for them, Releng signoff is fine).

I'll tweak the proposal slightly to reflect this.

mikeconley (commented):
This seems very reasonable to me. In my mind, at least for newtab, what'd be ideal is if one of the sign-offs comes from Product, and the other from an engineer - but for newtab, I imagine we'll invent our own internal process where our engineers will require written sign-off from QA and PM on the release before they will sign-off in ShipIt.

ahal:
We'd need another LDAP group for product managers in that case.. Maybe one product manager group for all system addons would be fine? We might also want to look into supporting people.mozilla.org groups in addition to LDAP so this is more self serve.

I think I'd propose we go ahead with the proposal as written first and then see how things shake out. We won't be locked into anything as a result of this RFC, and if there's a need for PM signoff in shipit, we can implement when it arises.

ahal:
Thanks all for the comments! Marking this Decided.

PR #55, opened 2026-04-16
RFC 0055: Replace Update Verify

bhearsum:
[rendered](https://github.com/bhearsum/releng-rfcs/blob/push-tywkvtwmqvmn/rfcs/0055-replace-update-verify.md)

### DIFF
--- a/rfcs/0055-replace-update-verify.md
+++ b/rfcs/0055-replace-update-verify.md
@@ -1,0 +1,113 @@
+# RFC 55 - Replace Update Verify
+* Comments: [#55](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/55)
+* Proposed by: @bhearsum
+
+# Summary
+
+Replace current "update verify" tests with modern, independent testing of MAR verification; get rid of the "liveness" test that it does entirely.
+
+## Background
+
+Update verify is a collection of bash and python scripts that verify aspects of the MARs that Firefox uses to bring itself up to date and the metadata in Balrog used to inform Firefox of which MAR to apply. These two checks are distinct, and described in more detail below.
+
+### MAR Verification
+
+The purpose of MAR verification checks is to ensure that MAR files get users to a functionally equivalent place as a fresh install. This is important for two reasons:
+1) It ensures that users get a functional build with the same code that automated and QE tests have been run against.
+2) It ensures that when users that apply this MAR update in the future, partial MARs will apply cleanly.
+
+### Liveness Checks
+
+The liveness check part of update verify ensures that the release metadata in Balrog contains entries for all necessary platforms, locale, and version combinations.
+
+Notably, this testing is very similar to what we do in the `final-verify` test.
+
+## Motivation
+
+There are a number of reasons to consider modernizing and changing the way we perform these tests:
+* Our current "update verify" tests represent a significant amount of time in the critical path of building releases. In betas, individuals tasks take close to an hour to complete and spent over 2 hours in end to end time to complete all update verify tasks. In releases, these numbers are 30 minutes and 1h45min. In both cases, the end to end time is a majority of the overall end to end time of the graph.
+* It is the least understood code that we own and is architecturally unsound: it is a cobbled together set of bash and python scripts that call back and forth to one another. For this and other reasons, it is difficult to confidently hack on.
+* Its output is poorly understood, often leading to delays in understanding and fixing failures.
+* Update verify tries to do too many things at once, which is partly to blame for the above two points.
+* In a review of its history, it has rarely caught issues that would hit users. In fact, it failed to catch [a very serious issue relatively recently](https://bugzilla.mozilla.org/show_bug.cgi?id=1890528).
+
+# Details
+
+## MAR Verification
+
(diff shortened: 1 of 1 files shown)


bhearsum:
This has been talked about elsewhere a bit, and no objections raised. With that in mind, I'm moving to final comment.

ahal (approved):
This all sounds great to me, thanks for writing it up!

rfcs/0055-replace-update-verify.md
```diff
@@ -0,0 +1,111 @@
...
+* It focuses exclusively on MAR verification.
+* It does not depend on Balrog to perform tests; this means tests can be run earlier in release graphs and are fully idempotent.
+* It runs tests in parallel, making it significantly faster than update verify for the same number of tests.
+
+In addition to using Marannon for running these tests, we should reduce the amount of testing that we do as follows:
+* Verify complete and partial MARS for all locales against `from` versions that have a partial generated for them. This ensures that all completes and partials are tested against recent versions.
+* Verify complete MARs for all locales against the first version after the last watershed for each platform. This does not test any MARs that weren't tested above, but it does verify the complete MARs against the most disparate version to the current one.
+
+The second point is a departure from `update-verify` in two ways:
+* It would eliminate all MAR verification done between those too points. It is noteworthy that over the nearly 20 years we have run update verify, we have no evidence of ever having an issue that would be caught by such checks.
```
hneiva: nit: too -> two

rfcs/0055-replace-update-verify.md
```diff
@@ -0,0 +1,111 @@
...
+
+The extra checks that update verify does are not necessary, and there is no evidence that they have ***ever*** caught a real issue:
+* The hashes returned by Balrog are not checked by Firefox itself anymore, so there is no point in doing so in our tests.
+* buildID, appVersion, and displayVersion are mostly just metadata for display, and the Balrog data and update verify configs ultimately get their information from the same place anyways.
+
+Checking that partials are smaller than completes is not covered at all by `final-verify`, which is an acceptable loss. If that check is something we desire to keep we should reimplement it to happen at partial MAR generation time instead.
+
+While we won't need to make any changes to `final-verify` itself as part of this proposal we will need to mark is as blocking `release-balrog-scheduling` to ensure that Balrog metadata is veirfied before we ship a release. (It currently blocks no other tasks; so failures could conceivably go unnoticed before shipping.)
```
hneiva: nit: verified

rfcs/0055-replace-update-verify.md
```diff
@@ -0,0 +1,111 @@
...
+
+# Open Questions
```
hneiva, line 101: Could we split the mar testing a bit more? Maybe multiple tasks per locale? (maybe this isn't the right place for this question)

also, is `update-verify-config` going away/being replaced? Is it worth mentioning here?

hneiva (commented):
🚀 Looks great!

rfcs/0055-replace-update-verify.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Proposed Testing
+
+We should not replicate this part of update verify. `final-verify` already performs the necessary checks (that all Balrog links return MARs, and that MARs are not 404s) on the live channel in the `push` phase of a release. The only advantage of doing a check like this earlier is that we would find issues with the Balrog metadata earlier. In practice, issues of this type virtually never happens these days.
+
+The extra checks that update verify does are not necessary, and there is no evidence that they have ***ever*** caught a real issue:
+* The hashes returned by Balrog are not checked by Firefox itself anymore, so there is no point in doing so in our tests.
+* buildID, appVersion, and displayVersion are mostly just metadata for display, and the Balrog data and update verify configs ultimately get their information from the same place anyways.
+
+Checking that partials are smaller than completes is not covered at all by `final-verify`, which is an acceptable loss. If that check is something we desire to keep we should reimplement it to happen at partial MAR generation time instead.
```
hneiva, line 97: We kind of already check for this:

https://searchfox.org/firefox-main/rev/410a52656dd13aa2360a3e96788e41bf48e80c0f/taskcluster/docker/partials-zucchini/make_incremental_zucchini.py#353-365

rfcs/0055-replace-update-verify.md
bhearsum: Splitting testing like that is out of scope here...but I will say that it's probably a net negative -- we're only running 5 tests (generally) in each task, and those are run in parallel, and depend on many of the same files. Splitting any further would mean we end up doing extra clones, extra downloads, and is unlikely to improve end to end times at all (it could even make them worse if the increased number of tasks causes tasks to pend longer).

As for `update-verify-config`, yes, it is no longer needed under this proposal and will be going away completely. That's worth calling out here, thanks for catching it!

rfcs/0055-replace-update-verify.md
bhearsum: That looks like it's at the patch-file-within-the-mar level? I guess that means it's unlikely that we could end up with a partial bigger than a MAR overall. This is a bit we can tackle if it comes up again in the future (which seems unlikely).

PR #56, opened 2026-05-19
RFC 0056: Improving RelEng Deployment Velocity & Confidence

bhearsum:
[rendered](https://github.com/bhearsum/releng-rfcs/blob/push-qxmrznsrlrmq/rfcs/0056-improve-velocity-and-confidence.md)

### DIFF
--- a/rfcs/0056-improve-velocity-and-confidence.md
+++ b/rfcs/0056-improve-velocity-and-confidence.md
@@ -1,0 +1,92 @@
+# RFC 0056 - Improving RelEng Deployment Velocity & Confidence
+* Comments: [#56](https://api.github.com/repos/mozilla-releng/releng-rfcs/issues/56)
+* Proposed by: @bhearsum
+
+# Summary
+
+We will speed up the rate at which we deploy changes to RelEng systems and improve our confidence around deployments. We will aim to deploy weekly and have automatic monitoring and rollback of deployments. This may pave the way for future improvements (eg: continuous deployments), but those are explicitly out of scope for the time being.
+
+## Background & Motivation
+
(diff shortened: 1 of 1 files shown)


ahal:
This is great, no notes!

Though I do have an adjacent thought. It would be neat if we could codify this in `mozilla-releng/reps`. E.g, have a separate template for "services" vs "packages".

I suspect most services will be different enough from one another that most of the things mentioned in this RFC will not be possible to make turnkey.. But perhaps there are some mozcloud config files, or taskcluster tasks that would generally be needed across any service we maintain.

This shouldn't be high priority because we basically never create new services these days, but on the other hand, `mozilla-releng/reps` now has the ability to update existing projects based on the template (ish)..

bhearsum:
> This is great, no notes!
> 
> Though I do have an adjacent thought. It would be neat if we could codify this in `mozilla-releng/reps`. E.g, have a separate template for "services" vs "packages".

This seems like a smart idea (although I wonder how it works out in a case like balrog, where we have a mini monorepo containing two services (`balrog` and `agent` and a package (`client`).

> 
> I suspect most services will be different enough from one another that most of the things mentioned in this RFC will not be possible to make turnkey.. But perhaps there are some mozcloud config files, or taskcluster tasks that would generally be needed across any service we maintain.

Most of the relevant MozCloud stuff lives in other repositories, but at the very least I guess we have a couple of GitHub actions that this could cover. (And IIRC, some of the Taskcluster IaC might come from the Taskcluster repo, so maybe there's a model for doing more?).

Taskcluster tasks could obviously be covered by this, of course.

jcristau:
I like this, thank you!

